PR #5858, opened 2026-03-07
fix: detect desktop runtime without frozen python

zouyonghe:
## Summary
- treat `ASTRBOT_DESKTOP_CLIENT=1` as the desktop backend runtime signal even when Python is not frozen
- restore desktop plugin installs to use `data/site-packages` and keep desktop root resolution aligned with the launcher contract
- add regression tests for runtime detection, explicit `ASTRBOT_ROOT` override, and `PipInstaller` target-site-packages behavior

## Test Plan
- [x] `.venv/bin/python -m pytest tests/test_runtime_env.py tests/test_pip_installer.py tests/test_dashboard.py -q`
- [x] `.venv/bin/ruff check astrbot/core/utils/runtime_env.py tests/test_runtime_env.py tests/test_pip_installer.py`

### DIFF
--- a/astrbot/core/utils/runtime_env.py
+++ b/astrbot/core/utils/runtime_env.py
@@ -7,4 +7,4 @@ def is_frozen_runtime() -> bool:
 
 
 def is_packaged_desktop_runtime() -> bool:
-    return is_frozen_runtime() and os.environ.get("ASTRBOT_DESKTOP_CLIENT") == "1"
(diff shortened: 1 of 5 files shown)


zouyonghe:
Superseded by #5859. This PR picked up unrelated history because  was rewritten upstream after I created the branch, so I rebuilt the change on top of the latest  and reopened it as a clean diff.

zouyonghe:
Closing in favor of #5859, which contains the same fix on top of the latest upstream  with a clean diff.

PR #6212, opened 2026-03-13
fix: handle all content items in multi-content tool results

### LINKED ISSUE #6211: Tool CallToolResult only processes first content item when multiple are returned
Hi there,

I noticed that when a tool returns a `CallToolResult` with multiple content items (e.g. both text and an image), only the first item gets processed. The rest are silently dropped.

Looking at `tool_loop_agent_runner.py`, the handler checks `res.content[0]` directly instead of iterating over all items in `res.content`. This means the LLM only sees partial results from tools that return rich/multi-part responses.

This seems related to #6140.

I have a fix ready — happy to open a PR if that works.

mango766:
Fixes #6211

When a tool returns a `CallToolResult` with multiple content items, only `content[0]` was being processed. This changes the handler to iterate over all items, so text, images, and embedded resources are all properly forwarded to the LLM.

**Changes:**
- Iterate over `res.content` instead of indexing `res.content[0]`
- Each item type (TextContent, ImageContent, EmbeddedResource) is handled correctly
- Empty content lists are handled gracefully
- Image index is tracked per-result so cached images get unique indices

**Testing:**
- Verified with tools that return single content items (no behavior change)
- Verified with tools that return multiple content items (all items now processed)

## Summary by Sourcery

Handle all content items returned from tool calls in the tool loop agent runner instead of only the first item.

Bug Fixes:
- Process all content entries from CallToolResult responses, including multiple text, image, and embedded resource items, instead of only the first.
- Gracefully handle tool results with empty content lists without failing.
- Assign unique indices to each cached image within a single tool result to avoid collisions.

### DIFF
--- a/astrbot/core/agent/runners/tool_loop_agent_runner.py
+++ b/astrbot/core/agent/runners/tool_loop_agent_runner.py
@@ -647,6 +647,33 @@ class ToolLoopAgentRunner(BaseAgentRunner[TContext]):
             async for resp in self.step():
                 yield resp
 
+    @staticmethod
+    def _cache_and_report_image(
+        base64_data: str,
+        func_tool_id: str,
+        func_tool_name: str,
+        image_index: int,
+        mime_type: str,
+    ) -> tuple[T.Any, str]:
+        """Cache an image and return (cached_img, description_text).
+
+        This is shared by both ImageContent and EmbeddedResource (blob image)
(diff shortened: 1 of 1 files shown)


Soulter:
Hi, #6149 already fixed this problem. Thanks for your PR!

PR #6401, opened 2026-03-15
fix: track handler task in session_waiter to prevent zombie tasks

rin259:
## Description

Fixes two issues:

1. **session_waiter.py**: Track handler task execution to prevent zombie tasks that continue running after session timeout
2. **gsv_selfhosted_source.py**: Update TODO comment for emotion analysis (not implemented)

### session_waiter.py
- Add  attribute to  class
- Cancel previous task before starting new one in  method
- Properly handle  when cancelling tasks

### gsv_selfhosted_source.py
- Replace TODO comment with clearer note that emotion analysis is not implemented

## Testing

Changes follow the existing code patterns and don't modify core logic.

## Summary by Sourcery

Prevent session handler zombie tasks by tracking and cancelling the active handler task, and clarify that emotion analysis is not yet implemented in the GSV self-hosted source.

Bug Fixes:
- Ensure only one session handler task runs at a time by cancelling any previously running handler task before starting a new one.

Enhancements:
- Track the currently running session handler task in the session waiter to support proper cancellation on timeout or retriggering.
- Clarify documentation in the self-hosted GSV source that emotion analysis is not implemented and must be integrated separately.

### DIFF
--- a/astrbot/core/provider/sources/gsv_selfhosted_source.py
+++ b/astrbot/core/provider/sources/gsv_selfhosted_source.py
@@ -141,7 +141,7 @@ class ProviderGSVTTS(TTSProvider):
         """
         params = self.default_params.copy()
         params["text"] = text
-        # TODO: 在此处添加情绪分析，例如 params["emotion"] = detect_emotion(text)
+        # 注意：情绪分析功能暂未实现，如需添加可接入情绪分析服务
         return params
 
     async def terminate(self) -> None:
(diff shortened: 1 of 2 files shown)


rin259:
Thanks for the feedback! 

Regarding the immediate await - we do need the task wrapper to enable cancellation of the previous handler. Here's why:

1. When a new trigger comes in while the previous handler is still running, we need to cancel that old task first
2. With plain , we'd have to wait for the previous handler to finish before starting the new one
3. Using  + immediate cancel allows us to interrupt the old handler and start fresh

I've added a  block to reset  after completion - this makes it clearer when no handler is running. Let me know if you'd like any other adjustments!

rin259:
Friendly ping — just checking whether there has been any progress on this PR. The fix is small and targeted; happy to adjust anything if needed. Thanks!

PR #6325, opened 2026-03-15
dev

LIghtJUNction:
This pull request introduces a new workflow for deploying the dashboard to GitHub Pages and makes significant improvements to the `README.md` for clarity, completeness, and consistency. It also includes minor formatting updates to the smoke test workflow and adds some convenience commands to `.envrc`.

**Summary of changes:**
- Adds a GitHub Actions workflow for automated dashboard deployment.
- Revamps the `README.md` with clearer descriptions, updated instructions, improved platform/model tables, and better contribution guidelines.
- Minor workflow formatting and comment changes for consistency.
- Resolves merge conflicts by merging the latest `master` changes into this branch.
- Hardens dashboard config selector path handling to prevent prototype-pollution key assignment (`__proto__`, `prototype`, `constructor`) in `dashboard/src/components/shared/AstrBotConfigV4.vue`.

---

**Dashboard Deployment Automation**

* Added `.github/workflows/deploy-dashboard.yml` to automate daily and manual dashboard builds and deployments to GitHub Pages, including build, artifact upload, and deployment steps.

**Documentation Improvements**

* Major rewrite and expansion of `README.md`: clearer project description, improved feature list, updated deployment instructions, revised supported platforms/models tables, and enhanced contribution guidelines. <a>[1]</a> <a>[2]</a> <a>[3]</a> <a>[4]</a> <a>[5]</a>

**Workflow Consistency**

* Updated `.github/workflows/smoke_test.yml` to use consistent YAML quoting, improved comments, and clarified Python version formatting. <a>[1]</a> <a>[2]</a> <a>[3]</a>

**Developer Convenience**

* Added `git pull` and `git status` commands to `.envrc` for easier environment setup and status checking.

**Post-feedback Updates**

* Resolved merge conflicts across backend, dashboard, i18n, and test files after syncing with `master`.
* Addressed a JavaScript CodeQL finding by adding guarded key checks in selector traversal/assignment logic in `AstrBotConfigV4.vue`.

PR #6404, opened 2026-03-15
feat: add OneBot V11 file API support for sending files

### LINKED ISSUE #6403: feat: add support for sending files via OneBot V11 file API
## Description

Currently, AstrBot's OneBot V11 adapter attempts to send files using / which doesn't properly support file sending. OneBot V11 provides dedicated APIs for file transmission:

-  - Send private file to user
-  - Send file to group

## Expected Behavior

When using  or  in a message chain, the bot should:

1. Use  or  for file messages
2. Pass the file path or URL as  parameter
3. Pass the display name as  parameter

## Current Issue

The current implementation in  treats File segments the same as regular messages and uses /, which doesn't properly handle file transmission.

## Implementation Suggestion

Modify  or create a separate handler for File segments to use the appropriate OneBot V11 file APIs:

This would be a significant improvement for users who need to transfer files through their QQ bots.

rin259:
## Description

Add support for sending files via OneBot V11's dedicated file APIs ( / ) instead of regular message API.

## Changes

Modified :

- When sending a  segment, now uses:
  -  for group messages
  -  for private messages
- Supports both local file paths () and URLs ()
- Added warning log for invalid session_id

## Related Issue

Fixes #6403

## Testing

This change uses the standard OneBot V11 file APIs which are supported by most OneBot V11 implementations (Go-CQHTTP, LLOneBot, etc.)

## Summary by Sourcery

Add support for sending file segments via OneBot V11 dedicated file APIs for group and private messages, with validation of target session identifiers.

New Features:
- Use OneBot V11 group and private file APIs when sending file segments instead of the generic message API.
- Allow file segments to be sent using either local file paths or URLs as the file source.

Bug Fixes:
- Prevent attempts to send files when the session_id is invalid by validating it before dispatching.

Enhancements:
- Log a warning when a file send is skipped due to an invalid session_id to aid in debugging message routing issues.

### DIFF
--- a/astrbot/core/platform/sources/aiocqhttp/aiocqhttp_message_event.py
+++ b/astrbot/core/platform/sources/aiocqhttp/aiocqhttp_message_event.py
@@ -179,7 +179,7 @@ class AiocqhttpMessageEvent(AstrMessageEvent):
                 else:
                     await bot.send_private_file(
                         user_id=session_id_int, file=file_path, name=file_name
-                    ) (refactor: improve file sending logic robustness)
+                    )
             else:
                 messages = await cls._parse_onebot_json(MessageChain([seg]))
                 if not messages:


rin259:
Thanks for the review! 

I've addressed the feedback in the latest commits:
- Added check for empty file_path before sending
- Fixed session_id check to handle ID 0 properly (using  instead of truthy check)
- Simplified the conditional logic

The changes are now more robust and handle edge cases correctly.

LIghtJUNction:
提交到dev将更快合并

LIghtJUNction:
你的源分支的基分支最好是dev分支，不然可能会导致冲突。

rin259:
> 你的源分支的基分支最好是dev分支，不然可能会导致冲突。

base branch已经是dev

LIghtJUNction:
@rin259 合并后到这里跟踪进度：https://github.com/AstrBotDevs/AstrBot/pull/6325

PR #6606, opened 2026-03-19
fix: prevent wecom ai bot long connection replies from disappearing

shijianhuai:
### Summary
#6585 This PR fixes an issue where Wecom AI Bot replies could disappear in long-connection mode after being displayed briefly in the WeCom client.

The root cause was that the pipeline scheduler always triggered a fallback empty send for WebChat and Wecom AI Bot events after pipeline execution, even when the event had already sent a valid response. In Wecom AI Bot long-connection mode, that extra fallback send produced an empty final stream message, which could overwrite the real reply on the client side.

### Root Cause
- command replies such as /help are sent through the normal send path instead of the streaming path
- after the response had already been sent, the scheduler still called send(None)
- in Wecom AI Bot long-connection mode, that fallback call became an extra empty final stream response
- the WeCom client then treated the empty final message as the latest result, making the previous visible reply disappear

### Changes
- only call the scheduler fallback send(None) when the event has not sent any response

### Impact
- fixes disappearing replies for Wecom AI Bot in long-connection mode
- keeps fallback behavior for cases where no response was sent at all

## Summary by Sourcery

Bug Fixes:
- Prevent Wecom AI Bot long-connection replies from being overwritten by an empty fallback message after a valid response was already sent.

### DIFF
--- a/astrbot/core/pipeline/scheduler.py
+++ b/astrbot/core/pipeline/scheduler.py
@@ -86,10 +86,9 @@ class PipelineScheduler:
         try:
             await self._process_stages(event)
 
-            # 如果没有发送操作, 则发送一个空消息, 以便于后续的处理
+            # 发送一个空消息, 以便于后续的处理
             if (
                 isinstance(event, WebChatMessageEvent | WecomAIBotMessageEvent)
-                and not event._has_send_oper
             ):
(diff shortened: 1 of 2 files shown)


Soulter:
used another better method to fix this bug.

PR #6715, opened 2026-03-20
fix(lark): support thread/reply-chain context isolation and reply_in_thread

nick2781:
## Summary

This PR adds proper thread and reply-chain support for the Lark (Feishu) platform adapter.

### Problem

- All messages in a group shared the same `session_id`, so conversations in different threads/reply-chains were mixed together.
- The bot always replied in the main chat even when the user messaged from within a thread or reply chain.
- `reply_in_thread` was hardcoded to `False`, preventing thread-aware replies.

### Changes

**`lark_adapter.py`**
- Build `session_id` with `%thread%{thread_id}` or `%root%{root_id}` suffix to isolate conversation context per thread/reply chain (both group and DM).
- Extract the real `chat_id` / `open_id` from `session_id` by stripping the `%thread%` / `%root%` suffix when sending active messages.
- Determine `is_in_thread` based on whether a thread needs to be **created** (`root_id` present but no `thread_id`), not whether the message is already in a thread.

**`lark_event.py`**
- Add `is_in_thread` attribute and `reply_in_thread` parameter through all send methods.
- When `reply_in_thread=True`, pass it to the Lark API `ReplyMessageRequestBody` so the reply creates/continues a thread.
- Fall back to non-streaming send when `is_in_thread=True`, because CardKit streaming cards do not support `reply_in_thread`.

### Key Design Decision

| Scenario | `thread_id` | `reply_in_thread` | Streaming |
|----------|-------------|-------------------|-----------|
| Message already in an existing thread | ✅ present | `False` (reply stays in thread naturally) | ✅ Works |
| Message in a reply chain, no thread yet | ❌ absent, `root_id` present | `True` (creates thread) | ❌ Falls back to non-streaming |
| Regular message (no thread/reply) | ❌ absent | `False` | ✅ Works |

### Testing

Tested in both group chat threads and reply chains on Feishu. Verified:
- Session isolation works correctly per thread/reply chain
- Streaming output works in existing threads
- Non-streaming fallback works when creating new threads
- Active message sending correctly extracts chat_id from suffixed session_id

## Summary by Sourcery

Add proper thread and reply-chain awareness to the Lark adapter so sessions and replies are isolated per thread and behave correctly in group and DM conversations.

Bug Fixes:
- Ensure Lark group and DM conversations no longer share context across different threads or reply chains by deriving distinct session IDs.
- Fix Lark replies being sent to the main chat instead of staying within or creating the appropriate thread or reply chain.
- Prevent use of streaming card responses when replying in thread contexts that do not support them, falling back to non-streaming sends instead.

Enhancements:
- Propagate an explicit thread-context flag through the Lark event and messaging pipeline to control reply_in_thread behavior for all message types (text, files, audio, media, cards).

### DIFF
--- a/astrbot/core/platform/sources/lark/lark_adapter.py
+++ b/astrbot/core/platform/sources/lark/lark_adapter.py
@@ -470,11 +470,17 @@ class LarkPlatformAdapter(Platform):
         if session.message_type == MessageType.GROUP_MESSAGE:
             id_type = "chat_id"
             receive_id = session.session_id
-            if "%" in receive_id:
+            # 从 session_id 中提取真实的 chat_id（去除 %thread% / %root% 后缀）
+            if "%thread%" in receive_id or "%root%" in receive_id:
+                receive_id = receive_id.split("%")[0]
+            elif "%" in receive_id:
                 receive_id = receive_id.split("%")[1]
         else:
             id_type = "open_id"
             receive_id = session.session_id
+            # 单聊也需要去除 %thread% / %root% 后缀
+            if "%thread%" in receive_id or "%root%" in receive_id:
+                receive_id = receive_id.split("%")[0]
 
         # 复用 LarkMessageEvent 中的通用发送逻辑
         await LarkMessageEvent.send_message_chain(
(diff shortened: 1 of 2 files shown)


nick2781:
Closing to test locally first — will reopen after verification.

PR #7468, opened 2026-04-11
fix: Telegram polling recovery after network failures

Shujakuinkuraudo:
## Summary
- rebuild the Telegram Application/Updater/HTTP client after polling network failures
- recover cleanly when the rebuild itself fails during initialization
- keep command registration lifecycle safe during shutdown/recovery
- add regression coverage for repeated polling failures and failed rebuild initialization

## Testing
- `.venv/bin/python -m ruff check astrbot/core/platform/sources/telegram/tg_adapter.py tests/test_telegram_adapter.py tests/fixtures/mocks/telegram.py`
- `.venv/bin/python -m pytest -q tests/test_telegram_adapter.py`

## Summary by Sourcery

Improve Telegram adapter resilience by rebuilding the application and client after repeated polling network failures and ensuring clean startup/shutdown during recovery and termination.

Bug Fixes:
- Restore Telegram polling after repeated network errors by rebuilding the Application/Updater/HTTP client once failures cross a configured threshold.
- Ensure Telegram adapter shutdown safely stops polling, application, and command registration without leaving dangling commands or updater state.
- Recover cleanly when a rebuilt Telegram application fails during initialization by shutting it down and constructing a fresh instance.

Enhancements:
- Track polling failure frequency and use an event-driven mechanism to trigger Telegram client rebuilds only after consecutive network errors within a time window.
- Refactor Telegram command registration scheduling into a dedicated helper that avoids duplicate scheduler startup and aligns with the new lifecycle management.

Tests:
- Add regression tests covering polling error thresholds, automatic application rebuild, skipping rebuild during termination, and recovery from rebuild initialization failures.
- Extend Telegram mocks with NetworkError/Forbidden/InvalidToken types and richer Application/Updater behavior to support the new polling recovery tests.

### DIFF
--- a/astrbot/core/platform/sources/telegram/tg_adapter.py
+++ b/astrbot/core/platform/sources/telegram/tg_adapter.py
@@ -3,12 +3,13 @@ import os
 import re
 import sys
 import uuid
+from contextlib import suppress
 from typing import cast
 
 from apscheduler.schedulers.asyncio import AsyncIOScheduler
 from telegram import BotCommand, Update
 from telegram.constants import ChatType
-from telegram.error import Forbidden, InvalidToken
+from telegram.error import Forbidden, InvalidToken, NetworkError
 from telegram.ext import ApplicationBuilder, ContextTypes, ExtBot, filters
 from telegram.ext import MessageHandler as TelegramMessageHandler
 
@@ -66,6 +67,7 @@ class TelegramPlatformAdapter(Platform):
             file_base_url = "https://api.telegram.org/file/bot"
 
         self.base_url = base_url
+        self.file_base_url = file_base_url
 
         self.enable_command_register = self.config.get(
             "telegram_command_register",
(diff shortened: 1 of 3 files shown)


Shujakuinkuraudo:
Follow-up note: this fix was also validated in a real Docker deployment against the failure mode that originally triggered the report.

Verified recovery chain:
- Telegram polling hits a proxy/network failure (`NetworkError` / `RemoteProtocolError`)
- the adapter rebuilds the Telegram `Application`/`Updater`/HTTP client
- if that rebuild attempt itself fails during `initialize()` (for example due to a transient timeout), the adapter discards the half-initialized instance and retries from a fresh application state
- polling can recover again once the network/proxy path becomes healthy

So the PR now covers both stages of the failure chain:
1. polling request failure
2. rebuild-initialization failure after recovery starts

All CI checks are green and the PR is now a single-commit change set.

Soulter (approved):

PR #7701, opened 2026-04-21
fix: apply provider modalities after tool injection

bugkeep:
## Summary
- move provider modality filtering until after built-in tool injection so models configured without tool_use do not receive late-added tools
- keep context sanitization aligned with the final provider/tool state
- add a regression test covering cron/proactive tools being stripped for text-only providers

## Verification
- confirmed the new regression test fails on current upstream before the fix
- uv run pytest tests/unit/test_astr_main_agent.py -q (80 passed)
- uv run ruff format .
- uv run ruff check .

Related to #6857.

## Summary by Sourcery

Adjust main agent build flow so provider modality filtering and context sanitization occur after all built-in tools are injected, ensuring text-only providers do not receive unsupported tools.

Bug Fixes:
- Ensure late-injected tools such as cron/proactive tools are stripped when the selected provider does not support tool use.

Tests:
- Add a regression test verifying that providers configured as text-only do not receive function tools after built-in tool injection.

### DIFF
--- a/astrbot/core/astr_main_agent.py
+++ b/astrbot/core/astr_main_agent.py
@@ -1424,10 +1424,8 @@ async def build_main_agent(
     if not req.session_id:
         req.session_id = event.unified_msg_origin
 
-    _modalities_fix(provider, req)
     _plugin_tool_fix(event, req)
     await _apply_web_search_tools(event, req, plugin_context)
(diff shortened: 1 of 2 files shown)


Soulter:
I put all modalities fix logic into tool loop agent runner in #7506 and fixed this issue

PR #7913, opened 2026-04-30
fix: deliver cron active agent fallback response

fxquarter:
## Summary
- add a fallback send path for active-agent cron jobs when the agent returns final assistant text but does not call `send_message_to_user`
- detect existing send-message tool output to avoid duplicate delivery
- cover fallback send and duplicate-skip behavior in cron manager tests

## Validation
- `ruff format .`
- `ruff check .`
- `uv run pytest tests/unit/test_cron_manager.py`

## Summary by Sourcery

Ensure active-agent cron jobs always deliver assistant responses by adding a fallback send path and avoiding duplicate deliveries when the agent already sent a message.

Bug Fixes:
- Add a fallback delivery path for active-agent cron jobs when the agent returns final assistant text without explicitly sending a user message.
- Prevent duplicate message delivery by detecting prior send-message tool outputs in provider request results.

Enhancements:
- Log detailed outcomes for active-agent cron fallback behavior, including skips, sends, and exceptions.

Tests:
- Add unit tests covering active-agent cron fallback delivery and the skip path when a message has already been sent.

### DIFF
--- a/astrbot/core/cron/manager.py
+++ b/astrbot/core/cron/manager.py
@@ -14,6 +14,7 @@ from astrbot.core.agent.tool import ToolSet
 from astrbot.core.cron.events import CronMessageEvent
 from astrbot.core.db import BaseDatabase
 from astrbot.core.db.po import CronJob
+from astrbot.core.message.message_event_result import MessageChain
 from astrbot.core.platform.message_session import MessageSession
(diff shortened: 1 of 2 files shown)


Soulter:
hi, thanks for your contribution, but not all cron jobs need to mention user actively

PR #7980, opened 2026-05-03
fix: normalize tool arguments using parameter schemas

whatevertogo:
Draft discussion PR for implementing a second-stage fix after #7961.

The goal is to centralize tool argument normalization so provider/model outputs like JSON-stringified arrays or objects can be recovered before tool execution, without adding one-off parsing blocks inside individual tools.

### Proposed direction

- Add a schema-guided normalizer for tool arguments.
- Coerce string values only when the tool parameter schema expects `array` or `object`.
- Support nested `properties`, `items`, `anyOf`, and `oneOf` conservatively.
- Do not parse parameters whose schema type is `string`, even if the string looks like JSON.
- Keep invalid JSON strings unchanged so existing tool validation still reports the error.

### Related cleanup

After the global path exists, the local `send_message_to_user.messages` JSON-string parsing workaround from #7978 can be removed and covered by shared normalizer tests.

### Scope notes

This should address issues similar to #7961 where the expected structured argument is delivered as a JSON string. It is not intended to solve unrelated tool-call problems such as duplicated tool names from streamed chunks or cases where a model already emitted `{}` and the original content is lost.

### Current status

This PR currently contains only an empty anchor commit so we can discuss the implementation shape before pushing code.

### Validation plan

- Unit tests for the normalizer:
  - array field from JSON string
  - object field from JSON string
  - nested object/list coercion
  - `anyOf` object/array coercion
  - string fields are not parsed
  - malformed JSON remains unchanged
- Integration-style test for `send_message_to_user.messages` going through the common normalization path.
- Existing targeted tests:
  - `uv run pytest tests/unit/test_message_tools.py`
  - targeted new normalizer tests
  - `uv run ruff check ...`
  - `uv run ruff format --check ...`


whatevertogo:
Closing this draft for now. The scoped #7961 fix remains in #7978, and the broader global tool-argument normalization does not look worth pursuing as a separate PR at this point.

PR #8058, opened 2026-05-07
fix(dashboard): merge plugin-registered providers into add-provider picker

Aluomolo:
## Summary

The **Add Provider** picker dialog calls `GET /api/config/provider/template`, which previously returned only the static `CONFIG_METADATA_2`. Plugin-registered providers (via `@register_provider_adapter`) live in `provider_registry`; that registry was already merged into the response of `GET /api/config/get` (see `_get_astrbot_config`, [`config.py:1481-1487`](https://github.com/AstrBotDevs/AstrBot/blob/master/astrbot/dashboard/routes/config.py#L1481-L1487)), but not here. As a result, plugin TTS / STT / LLM providers never appear in the picker even though they are present in `provider_registry`.

This patch mirrors the same merge in `get_provider_template` so the two endpoints are consistent.

## Repro (before fix)

1. Write a plugin that registers a custom TTS provider:
   ```python
   from astrbot.core.provider.entities import ProviderType
   from astrbot.core.provider.provider import TTSProvider
   from astrbot.core.provider.register import register_provider_adapter

   @register_provider_adapter(
       'my_custom_tts', 'My Custom TTS',
       provider_type=ProviderType.TEXT_TO_SPEECH,
       default_config_tmpl={'provider_type': 'text_to_speech', 'api_key': ''},
   )
   class MyCustomTTS(TTSProvider):
       async def get_audio(self, text: str) -> str: ...
   ```

2. Start AstrBot, confirm provider is registered (debug log: `Model provider registered: my_custom_tts`).

3. Compare the two endpoints:
   - `GET /api/config/get` → entry present at `.data.metadata.provider_group.metadata.provider.config_template.my_custom_tts` ✅
   - `GET /api/config/provider/template` → entry **missing** at `.data.config_schema.provider.config_template` ❌

4. WebUI → 服务提供商 → 添加 → the custom provider does not appear in any tab.

## After fix

`get_provider_template` performs the same merge, so both endpoints expose plugin providers. Picker now displays them in the appropriate tab (filtered by `provider_type`).

## Why no test

The existing dashboard route tests in `tests/` only cover pure helper functions (e.g. `tests/test_chat_route.py` tests `_poll_webchat_stream_result`). A route-level test for `get_provider_template` would require constructing the full `ConfigRoute` with `core_lifecycle` / DB / Quart fixtures — out of proportion to the 5-line fix. The change mirrors existing code in the same file (`_get_astrbot_config`, lines 1481-1487), so it is mechanical and easy to review.

## Test Plan

- [x] `from astrbot.dashboard.routes.config import ConfigRoute` imports cleanly
- [x] Manually verified locally: with a plugin registering a TTS provider, picker now shows the entry
- [ ] CI checks

## Summary by Sourcery

Bug Fixes:
- Ensure plugin-registered provider adapters appear in the WebUI Add Provider picker by injecting their config templates into the provider template response.

### DIFF
--- a/astrbot/dashboard/routes/config.py
+++ b/astrbot/dashboard/routes/config.py
@@ -330,6 +330,28 @@ def save_config(
     config.save_config(post_config)
 
 
+def _merge_registered_providers_into(config_template: dict) -> None:
+    """Inject providers registered via ``@register_provider_adapter`` into
+    a config_template dict, in-place.
+
+    Used by both ``GET /api/config/get`` and ``GET /api/config/provider/template``
+    so the two endpoints expose a consistent set of providers in the WebUI's
+    "Add Provider" picker.
+
+    - Uses ``is not None`` (not truthiness) so providers that intentionally
+      register an empty default template still appear.
+    - Uses ``setdefault`` so a plugin cannot silently shadow a core static
+      template that happens to share the same key.
+
+    The caller owns ``config_template`` and is responsible for handing in a
+    non-shared dict (both call sites operate on already-deep-copied metadata
+    so mutating it here does not pollute ``CONFIG_METADATA_2``).
+    """
+    for provider in provider_registry:
+        if provider.default_config_tmpl is not None:
+            config_template.setdefault(provider.type, provider.default_config_tmpl)
+
+
 class ConfigRoute(Route):
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


Aluomolo:
Thanks for the reviews — addressed all three points in the amended commit (force-pushed):

1. **`is not None` vs truthiness** (sourcery): switched to `is not None` so plugins that register with an intentionally empty `default_config_tmpl={}` still appear in the picker.

2. **`setdefault` to protect core entries** (sourcery): used `dict.setdefault` so a plugin cannot silently shadow a core static template that happens to share the same key.

3. **Shared helper** (sourcery + gemini-code-assist): extracted the merge into a module-level `_merge_registered_providers_into()` and routed both `_get_astrbot_config` and `get_provider_template` through it, so the two endpoints can't drift.

4. **Avoid mutating global state** (gemini-code-assist): `get_provider_template` now `copy.deepcopy`s the static `provider` section from `CONFIG_METADATA_2` before merging, so plugin entries no longer leak into the global dict across requests. (`_get_astrbot_config` was already operating on a deepcopy of the whole metadata, so it didn't need a change there.)

Added `tests/unit/test_provider_template_merge.py` covering injection, setdefault collision protection, empty-vs-None template distinction, and idempotence — 6 tests, all green. Full unit suite still passes apart from two pre-existing `test_computer.py` failures unrelated to this change.

PR #8063, opened 2026-05-07
feat(whisper): expose language + prompt config (multilingual opt-in)

YonganZhang:
## Problem

`ProviderOpenAIWhisperAPI` does not forward `language` / `prompt` parameters to `client.audio.transcriptions.create()`. Whisper auto-detect can mis-classify non-English speech (Chinese / Japanese / Korean / etc), hurting transcription accuracy. Users have no way to hint the language or provide a prompt through the existing provider config.

## Solution

Expose two optional config fields, both defaulting to `""` (preserves current auto-detect behavior — fully backwards compatible):

- `language`: ISO language hint, e.g. `"zh"` / `"ja"` / `"ko"`
- `prompt`: free-text guidance ([OpenAI prompting guide](https://platform.openai.com/docs/guides/speech-to-text/prompting))

Plus `temperature=0` for deterministic transcription output.

## Backwards compatibility

Default `""` → `NOT_GIVEN` → Whisper auto-detect → identical behavior to current code. Existing users see no change. New parameters are pure opt-in.

## Test

Tested locally on Chinese voice samples (PolyU lab). Setting `language="zh"` plus a short Chinese context prompt produced visibly better WER than the previous auto-detect path.

## Diff size

8 effective lines: 5 init + 3 transcription call.

## Summary by Sourcery

Expose configurable language and prompt options for the Whisper API provider to improve non-English transcription accuracy while preserving existing auto-detection behavior.

New Features:
- Add optional language and prompt configuration fields to the Whisper API provider for guiding transcriptions.

Enhancements:
- Forward configured language and prompt to the Whisper transcription call and set temperature to 0 for more deterministic output.


Soulter (commented):
it's better to surface these configs in astrbot config

YonganZhang:
Addressed the four review findings from `@gemini-code-assist`:

| # | Finding | Fix |
|---|---|---|
| 1 | `open(audio_url, "rb")` leaks the file handle; subsequent `os.remove(audio_url)` can fail with EBUSY on Windows and accumulate FDs on POSIX under concurrency | Read bytes inside a `with` block, pass `audio_bytes` to `transcriptions.create`. Handle is released before `os.remove`. |
| 2 | `language` / `prompt` used verbatim from YAML — accidental whitespace would change transcription | `(provider_config.get(...) or "").strip()` — also accepts `None` cleanly |
| 3 | `temperature=0` hardcoded | New optional `provider_config["temperature"]` (default `0`), coerced via `float()` with `0.0` fallback for non-numeric values |
| 4 | Missing unit tests | Left as follow-up — `tests/` doesn't yet have a whisper-specific harness and adding one would expand scope. Happy to open a separate `tests/test_whisper_api_source.py` PR once maintainers confirm preferred mocking style (httpx_mock vs OpenAI client mock). |

`@gemini-code-assist review`

YonganZhang:
Correction on point 4: `tests/test_whisper_api_source.py` already exists on the current PR HEAD (added upstream via commit `8098a92`). The bot's "missing unit tests" finding appears to be a false positive — its scan must have predated that commit. The existing test exercises `ProviderOpenAIWhisperAPI` with a mocked OpenAI client and should keep passing under this PR's changes:

* `self.language = (provider_config.get("language") or "").strip()` — unchanged for the test's `provider_config` (no `language` key, so both old and new resolve to `""`).
* `self.temperature` — defaults to `0.0` when not configured (matches the previous hard-coded `0`).
* File-handle change is internal to `transcribe()` — the test's mocked `create` doesn't inspect the `file=` argument shape.

So no test changes needed, and the existing harness already covers the call-site contract. Happy to add focused cases for the new `temperature` / `strip()` behavior in a follow-up if maintainers want explicit coverage.

YonganZhang:
Self-correction on the file-handle fix from a few minutes ago:

I had switched `file=("audio.wav", open(...))` to pass raw `bytes`, but the existing `tests/test_whisper_api_source.py` asserts on `file_arg[1].name.endswith(".wav")` and calls `file_arg[1].close()` — both require a real file-like object. The bytes form would have broken the test.

Reverted to passing the file handle, with an **explicit `try / finally close()`**:

```python
audio_file = open(audio_url, "rb")
try:
    result = await self.client.audio.transcriptions.create(
        ...,
        file=("audio.wav", audio_file),
        ...,
    )
finally:
    audio_file.close()
```

The handle is now guaranteed to be released **before** the subsequent `os.remove(audio_url)` — which was the actual leak `@gemini-code-assist` flagged — and the SDK contract is unchanged so the existing test keeps passing.

YonganZhang:
Addressed three of `@gemini-code-assist`'s new findings:

**HIGH: downloaded temp file leak** ✅
When `audio_url` is an http(s) URL we download into `whisper_api_<uuid>.input` and reassign `audio_url` to that local path. If the downloaded format **does not** need conversion (mp3 / wav / m4a / mpeg / webm — i.e., not opus/silk/amr), `output_path` stays `None` and the previous cleanup `if output_path and os.path.exists(...)` **never touched the downloaded file** — it accumulated in `temp_dir` on every URL transcription.

Fix: track `downloaded_path` separately, walk both `(output_path, downloaded_path)` in the cleanup loop. Identical no-op when only one is set.

**MED: `str()` cast on config values** ✅
`(provider_config.get("language") or "").strip()` raises `AttributeError` if a YAML editor accidentally types an int or bool. Cast to `str` before `strip()` so the init is robust against non-string config values.

**MED: stream audio directly / move await inside with-block** — the current `try / finally close()` around the SDK call is behavior-equivalent (handle released before `os.remove`, no `audio_file.read()` materialization). Leaving the shape as-is keeps the existing `tests/test_whisper_api_source.py` assertions on `.name` and `.close()` valid.

**MED: unit tests for language/prompt extraction** — `tests/test_whisper_api_source.py` already exists (added upstream in commit `8098a92`); happy to add a focused case for the new `temperature` / `strip()` / leak-fix behavior as a follow-up PR rather than expand this scope.

`@gemini-code-assist review`

PR #8064, opened 2026-05-07
fix(minimax-tts): surface real API error when response is not SSE

YonganZhang:
## Bug

MiniMax TTS API returns a JSON error body (not an SSE stream) when:
- quota / rate limit exceeded
- invalid `voice_id` or model
- API key issues

Current code calls `raise_for_status()` (which doesn't trigger on `200 OK` with a JSON error body) and then parses everything as SSE. Users get a confusing parse error or empty audio with no indication of the real cause.

## Fix

Before entering the SSE parse loop, check `Content-Type`. If it's not `text/event-stream`, parse the JSON `base_resp` and raise a `RuntimeError` carrying the actual `status_code` and `status_msg`.

## Example

**Before:**
```
TTS failed: Cannot decode SSE event from b'{"base_resp":{"status_code":1027,"status_msg":"insufficient balance"}}'
```

**After:**
```
RuntimeError: MiniMax TTS API error (code=1027): insufficient balance
```

## Test

Reproduced locally with a depleted MiniMax API key — error message now clearly identifies the cause; prior version produced a confusing SSE parse error.

## Diff size

17 lines, pure error-handling. No new feature, no new dependency.

## Summary by Sourcery

Bug Fixes:
- Detect non-SSE MiniMax TTS responses via Content-Type and raise a RuntimeError with the API's status_code and status_msg instead of producing a confusing SSE parse failure.

### DIFF
--- a/astrbot/core/provider/sources/minimax_tts_api_source.py
+++ b/astrbot/core/provider/sources/minimax_tts_api_source.py
@@ -111,17 +111,26 @@ class ProviderMiniMaxTTSAPI(TTSProvider):
                 # Detect by Content-Type and surface the actual API error so callers
                 # see the real status_code / status_msg instead of a confusing
                 # parsing failure later in the SSE loop.
-                content_type = response.headers.get("Content-Type", "")
-                if "text/event-stream" not in content_type:
+                # MIME types are case-insensitive (RFC 7231 §3.1.1.1), and some
+                # error responses include parameters like "application/json; charset=utf-8".
+                # Lower-case the value and compare against the bare media-type prefix.
+                content_type = response.headers.get("Content-Type", "").lower().split(";", 1)[0].strip()
+                if content_type != "text/event-stream":
                     body = await response.text()
+                    err_msg = body[:200] or "empty response body"
+                    err_code = "unknown"
                     try:
                         err_data = json.loads(body)
-                        base_resp = err_data.get("base_resp", {})
-                        err_msg = base_resp.get("status_msg", body[:200])
-                        err_code = base_resp.get("status_code", "unknown")
+                        # Guard against `base_resp: null`, missing key, or a JSON
+                        # array root — all of which would have raised
+                        # AttributeError on `.get(...)` before this change.
+                        if isinstance(err_data, dict):
+                            base_resp = err_data.get("base_resp")
+                            if isinstance(base_resp, dict):
+                                err_msg = base_resp.get("status_msg", err_msg)
+                                err_code = base_resp.get("status_code", err_code)
                     except json.JSONDecodeError:
-                        err_msg = body[:200]
-                        err_code = "unknown"
+                        pass
                     raise RuntimeError(
                         f"MiniMax TTS API error (code={err_code}): {err_msg}"
                     )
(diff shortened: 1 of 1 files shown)


YonganZhang:
Addressed both review findings.

**`@sourcery-ai` — Normalize Content-Type before checking** ✅
Now lower-cased and split on `;` to strip any `charset=...` suffix. Both `Text/Event-Stream` (mixed case) and `text/event-stream; charset=utf-8` (with parameter) are correctly recognized as the SSE success path.

**`@gemini-code-assist` — Fragile JSON parsing** ✅
Guarded with `isinstance(...)` checks on both `err_data` (could be a JSON array root) and `base_resp` (could be JSON `null` or missing). Fallback values for `err_msg` / `err_code` are computed up-front so an inner-path exception can no longer leave them unbound.

### Trip-test (every previously-crashing input now produces a clean RuntimeError)

| Input | Before | After |
|---|---|---|
| `text/event-stream` body | OK | OK (unchanged) |
| `Text/Event-Stream` (mixed case) | misclassified as error 🔴 | OK ✅ |
| `text/event-stream; charset=utf-8` | misclassified as error 🔴 | OK ✅ |
| JSON with `{"base_resp": {...}}` | `MiniMax TTS API error (code=1004): quota exceeded` | unchanged ✅ |
| JSON with `{"base_resp": null}` | **AttributeError** 🔴 | `MiniMax TTS API error (code=unknown): ...` ✅ |
| JSON array root `[{"err":1}]` | **AttributeError** 🔴 | `MiniMax TTS API error (code=unknown): ...` ✅ |
| JSON missing `base_resp` | `code=unknown` ✅ | unchanged ✅ |
| Non-JSON body | `code=unknown` ✅ | unchanged ✅ |
| Empty body | `code=unknown: ""` (sparse) | `code=unknown: empty response body` ✅ |

`@sourcery-ai review` `@gemini-code-assist review`

YonganZhang:
Addressed `@gemini-code-assist`'s reorder suggestion + clarified the `RuntimeError` vs `Exception` choice.

**1. Move Content-Type check before `raise_for_status()`** ✅
4xx responses from MiniMax *do* carry the structured `base_resp` body for quota / rate-limit / invalid voice failures — we want to surface `base_resp.status_code / status_msg` on those instead of an opaque `aiohttp.ClientResponseError`. Behavior matrix after this PR:

| status | Content-Type | result |
|---|---|---|
| 200 | text/event-stream | SSE loop (success, unchanged) |
| 200 | application/json | `RuntimeError` with `base_resp` (unchanged) |
| **4xx** | **application/json** | **`RuntimeError` with `base_resp` (NEW — was opaque)** |
| 5xx | text/html | `raise_for_status` raises (unchanged) |
| 4xx | empty body | `RuntimeError "empty response body"` → `raise_for_status` fallback |

`response.raise_for_status()` stays at the end of the discrimination block as a fallback — non-2xx responses that are *neither* SSE *nor* a structured JSON error still surface as transport errors, which is the right signal for operators chasing 5xx outages.

**2. `Exception` vs `RuntimeError`** — keeping `RuntimeError` because the rest of `astrbot/core/provider/sources/*` uses `RuntimeError` for application-level provider failures. Switching to bare `Exception` here would be inconsistent with neighboring providers.

**3. Unit tests** — same scope-deferral as #8063: happy to follow up with a focused aiohttp-mock harness once maintainers confirm the preferred style.

`@gemini-code-assist review`

PR #8068, opened 2026-05-07
feat(knowledge-base): add `enabled` flag for soft enable/disable

YonganZhang:
## Use case

Currently the only way to "turn off" a knowledge base is to delete it, which loses all indexed documents and embeddings. Users frequently want to temporarily disable a KB (e.g. testing, debugging, A/B comparison) without paying the re-indexing cost later.

## Change

Three small additions to the existing schema:

1. `KnowledgeBase` model gets an `enabled: bool = True` field.
2. `kb_db_sqlite.migrate_to_v2()` adds the column to existing DBs via `ALTER TABLE ... ADD COLUMN enabled BOOLEAN NOT NULL DEFAULT 1`, wrapped in `try/except` so re-running the migration is a no-op.
3. `RetrievalManager.invalidate_sparse_cache(kb_id)` exposes a small helper to clear the BM25 cache when a KB is enabled/disabled or its docs change. Useful for the dashboard toggle and any other consumer that mutates KB state.

## Backwards compatibility

- New rows default to `enabled=True` → no behavior change for existing KBs.
- Existing rows get `1` (true) via the migration default → no behavior change.
- Pure additive: no existing field renamed or removed.

## Diff size

16 lines across 3 files. Pure model + migration + helper — no API contract change in this PR.

## Follow-up

A separate PR can wire this up in the dashboard (toggle button + filter `enabled=true` in retrieval). Posting this as a small, self-contained schema change first to keep review focused.

## Summary by Sourcery

Introduce a soft enable/disable mechanism for knowledge bases and wire supporting schema migration and cache invalidation helper methods.

New Features:
- Add an enabled flag to knowledge bases to support soft enable/disable behavior.
- Expose a retrieval manager helper to invalidate BM25 sparse cache for a specific knowledge base.

Enhancements:
- Add a v2 SQLite migration to backfill the enabled column on existing knowledge base rows without impacting current behavior.

### DIFF
--- a/astrbot/core/knowledge_base/kb_db_sqlite.py
+++ b/astrbot/core/knowledge_base/kb_db_sqlite.py
@@ -78,12 +78,6 @@ class KBSQLiteDatabase:
             await conn.execute(text("PRAGMA optimize"))
             await conn.commit()
 
-        # v2 schema migration: add the `enabled` column to knowledge_bases.
-        # Idempotent and re-entrant — safe to call on every startup; the
-        # method swallows the "duplicate column" path quietly and logs anything
-        # else at debug. See migrate_to_v2 docstring.
-        await self.migrate_to_v2()
-
         self.inited = True
 
     async def migrate_to_v1(self) -> None:
@@ -192,7 +186,11 @@ class KBSQLiteDatabase:
                 if "duplicate column" in msg or "already exists" in msg:
                     # Column already present from a prior migration run — expected.
                     return
-                logger.debug(f"知识库 v2 迁移跳过 (非重复列错误): {e!r}")
+                # Real schema failure — let it propagate so the manager surfaces
+                # a broken-knowledge-base state on startup instead of silently
+                # running with the old schema.
+                logger.error(f"知识库 v2 迁移失败: {e!r}")
+                raise
 
     async def close(self) -> None:
         """关闭数据库连接"""
--- a/astrbot/core/knowledge_base/kb_db_sqlite.py
+++ b/astrbot/core/knowledge_base/kb_db_sqlite.py
@@ -78,6 +78,12 @@ class KBSQLiteDatabase:
             await conn.execute(text("PRAGMA optimize"))
             await conn.commit()
 
+        # v2 schema migration: add the `enabled` column to knowledge_bases.
+        # Idempotent and re-entrant — safe to call on every startup; the
+        # method swallows the "duplicate column" path quietly and logs anything
+        # else at debug. See migrate_to_v2 docstring.
+        await self.migrate_to_v2()
+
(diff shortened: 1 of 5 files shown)


YonganZhang:
Addressed the bot review findings in commit ff8f65f:

**HIGH bugs (both shipped in this update):**

1. **`migrate_to_v2` was never called** (`@gemini-code-assist`) — Wired it into `KnowledgeBaseDBSqlite.initialize()` right after the `engine.begin()` PRAGMA block commits, so existing databases pick up the new `enabled` column on startup. The migration is idempotent — subsequent runs see "duplicate column" and quietly return.

2. **`SparseRetriever.invalidate_cache` was undefined** (`@gemini-code-assist`) — `RetrievalManager.invalidate_sparse_cache` would have raised `AttributeError` at runtime. Added `SparseRetriever.invalidate_cache(kb_id)` that pops the entry from the internal `_index_cache` dict (no-op if absent).

**MEDIUM (also fixed):**

3. **`migrate_to_v2` swallowed all exceptions silently** (`@sourcery-ai` / `@gemini-code-assist`) — Narrowed the swallow: only the SQLite "duplicate column" / "already exists" idempotency path returns silently; every other exception is logged at `debug` so a real schema problem is no longer invisible.

`@gemini-code-assist review` `@sourcery-ai review`

YonganZhang:
Addressed both new findings from `@gemini-code-assist`'s re-review:

**1. Migration order (v2 was running before v1)** ✅
`migrate_to_v1` is invoked from `kb_mgr.py:61` (manager layer), not from `KnowledgeBaseDBSqlite.initialize()`. My previous commit's `await self.migrate_to_v2()` inside `initialize()` therefore ran **before** `migrate_to_v1`, which is the wrong order — `migrate_to_v1` creates indices that `migrate_to_v2`'s `enabled` column should sit on top of. Removed the call from `initialize()`, and added `await self.kb_db.migrate_to_v2()` immediately after `migrate_to_v1()` in `kb_mgr.py` so the v1 → v2 order is the same as the function naming.

**2. Re-raise unexpected migration errors** ✅
Previous commit caught everything-not-"duplicate column" and only `logger.debug`'d. A real schema failure (DB locked, missing table, permission error) silently left the app running on the old schema. Switched the non-idempotent path to `logger.error` + `raise` so the manager surfaces the broken-KB state on startup instead of running with stale schema.

**3. Unit tests** — agree these would help; the existing `tests/` tree doesn't yet have a KB-DB harness, so adding one would expand this PR's scope. Happy to follow up with a focused `tests/test_kb_db_sqlite.py` PR once maintainers confirm the preferred mocking style (in-memory SQLite vs full integration).

`@gemini-code-assist review`

PR #8074, opened 2026-05-08
feat(cua): expire idle sandbox sessions

zouyonghe:
## Summary
- add a CUA-only `cua_idle_timeout` config that proactively shuts down idle sandbox sessions
- track one idle cleanup task per CUA session and reset the timer whenever the session is reused
- add regression tests covering proactive expiration, timer refresh on reuse, disabled timeout behavior, and non-CUA isolation

## Problem
AstrBot caches sandbox booters per session in memory. For CUA sessions this means a local sandbox can stay alive indefinitely unless the process exits, the booter becomes unavailable, or the user explicitly asks the agent to close it. In practice that leaves local Docker-backed CUA sandboxes running far longer than users expect.

## Root Cause
`computer_client.get_booter()` stores active sandbox booters in `session_booter` and reuses them for the same session. The CUA booter itself uses an ephemeral sandbox context, but AstrBot does not currently release that context when the session becomes idle, so the sandbox lifetime ends up tied to process lifetime rather than actual usage.

## Fix
This change adds a CUA-specific proactive idle timeout.

- New config: `provider_settings.sandbox.cua_idle_timeout`
- Default: `0` (disabled, preserves current behavior)
- When enabled, AstrBot creates one async cleanup task per active CUA session
- The task sleeps until the remaining idle budget is exhausted, then calls `booter.shutdown()` and removes the session from the in-memory cache
- Reusing the same CUA session resets the timer by cancelling the previous cleanup task and scheduling a fresh one

The implementation is intentionally scoped to CUA only so other sandbox booters keep their current lifecycle behavior.

## Validation
- `uv run pytest tests/unit/test_cua_computer_use.py tests/unit/test_computer.py -q`
- `uv run ruff check astrbot/core/computer/computer_client.py astrbot/core/config/default.py tests/unit/test_cua_computer_use.py`

## Summary by Sourcery

Add configurable idle-timeout handling for CUA sandbox sessions so idle sessions are proactively shut down while preserving existing behavior by default.

New Features:
- Introduce a CUA-only configuration option to control sandbox idle timeout and default it to disabled.
- Track per-session idle state for CUA sandbox booters and schedule asynchronous cleanup when sessions exceed the configured idle timeout.

Enhancements:
- Ensure CUA idle timers are refreshed when an existing session is reused and cleared on shutdown or booter errors.
- Keep non-CUA sandbox booters unaffected by the new idle-timeout behavior.

Tests:
- Add regression tests covering proactive CUA session shutdown on idle, timer refresh on reuse, disabled-timeout behavior, and isolation from non-CUA booters.

### DIFF
--- a/astrbot/core/computer/computer_client.py
+++ b/astrbot/core/computer/computer_client.py
@@ -4,6 +4,7 @@ import os
 import shutil
 import time
 import uuid
+from dataclasses import dataclass
 from pathlib import Path
 
 from astrbot.api import logger
@@ -19,9 +20,16 @@ from .booters.local import LocalBooter
 
 session_booter: dict[str, ComputerBooter] = {}
 local_booter: ComputerBooter | None = None
-cua_last_used_at: dict[str, float] = {}
-cua_idle_cleanup_tasks: dict[str, asyncio.Task] = {}
 _MANAGED_SKILLS_FILE = ".astrbot_managed_skills.json"
+
+
+@dataclass(slots=True)
+class _CUAIdleState:
+    expires_at: float
+    task: asyncio.Task
+
+
+cua_idle_state: dict[str, _CUAIdleState] = {}
 
 
 def _get_cua_idle_timeout(config: dict) -> float:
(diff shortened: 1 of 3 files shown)


zouyonghe:
@sourcery-ai review

PR #8155, opened 2026-05-11
fix(provider/openai): normalize oversized tool_call IDs to avoid 400 from strict upstreams

x1051445024:
# Normalize oversized `tool_call` IDs to avoid 400 from strict upstreams

## Problem

Some OpenAI-compatible relay services (e.g. "溏心", oneapi-like proxies) return
`tool_call.id` strings that far exceed the 64-character limit enforced by the
OpenAI API spec. Observed lengths in production traffic: **660 and 1650+ characters**.

Because AstrBot round-trips those IDs back into the next request's
`messages[].tool_calls[].id` / `tool_call_id` fields, the upstream responds with:

```
openai.BadRequestError: Error code: 400 - {
  'error': {
    'message': "Invalid 'input[16].call_id': string too long. Expected a string
                with maximum length 64, but got a string with length 1650 instead.",
    'type': 'invalid_request_error',
    'param': 'input[16].call_id',
    'code': 'string_above_max_length'
  }
}
```

Some relays internally translate Chat Completions payloads into the Responses API
format, which renames `tool_call_id` to `call_id` — but the root cause is the
same: the ID AstrBot emits is too long to begin with.

## Fix

Two layers of defense in `core/provider/sources/openai_source.py`:

### 1. Response-parsing layer

In `_parse_openai_completion`, any freshly-received `tool_call.id` longer than
64 chars is deterministically shortened to `call_<md5>` (37 chars) before it
enters the conversation history. The associated `extra_content` (used by
Gemini-2.5/3 pass-through) is re-keyed with the normalized ID.

### 2. Pre-send fallback layer

A new static method `_normalize_tool_call_ids` scans
`payloads["messages"]` before every Chat Completions request (both `_query` and
`_query_stream`). It collects every oversized ID across assistant
`tool_calls[].id` and tool `tool_call_id` into a shared map, then rewrites them
in a second pass. This guarantees assistant-call / tool-result pairing stays
intact, and also rescues long IDs persisted in the history before the fix was
deployed.

Shortening is deterministic (MD5 of the original ID), so a single relay session
stays internally consistent across retries. No new dependencies are introduced
— `hashlib` is already imported at the top of the file.

## Verification

- Before: every tool-using request against the affected relay returned HTTP 400.
- After: request succeeds; log shows
  `[tool_call_id 规范化] 发送前兜底：共 N 个超长 ID 被短化` when oversized
  IDs are encountered, and nothing when they are not.

## Risk assessment

- **Scope**: limited to `openai_source.py`; only executes on oversized IDs.
- **Backwards compatibility**: when all IDs are already ≤64 chars, both layers
  are no-ops (the fallback method returns early on an empty map).
- **Provider coverage**: only Chat Completions path (OpenAI-compatible). Anthropic
  and Gemini native sources are untouched — they have their own ID schemes.

## Summary by Sourcery

Normalize and sanitize oversized OpenAI tool_call IDs to prevent upstream 400 errors and keep assistant/tool message IDs consistent.

Bug Fixes:
- Shorten overlong tool_call IDs in parsed OpenAI chat completions to stay within the 64-character API limit.
- Normalize oversized tool_call IDs in outgoing chat completion payloads so assistant tool_calls and tool responses remain correctly paired without triggering validation errors.

Enhancements:
- Add warning logs when tool_call IDs are normalized to aid in diagnosing relay or provider issues.

### DIFF
--- a/astrbot/core/provider/sources/openai_source.py
+++ b/astrbot/core/provider/sources/openai_source.py
@@ -564,19 +564,39 @@ class ProviderOpenAIOfficial(Provider):
 
         payloads["messages"] = cleaned
 
-     @staticmethod
+    @staticmethod
+    def _shorten_tool_call_id(raw_id: str | None) -> str | None:
+        """Deterministically shorten an oversized tool_call ID.
+
+        Non-cryptographic by design; MUST NOT be used for any
+        security-sensitive purpose. Its only job is to normalize IDs
+        that exceed the 64-character limit enforced by the OpenAI API
+        spec into a stable, compact form so the same ID collapses to
+        the same short form across retries of the same request.
+
+        Short IDs (or empty/None) are returned unchanged.
+        """
+        if not raw_id or len(raw_id) <= 64:
+            return raw_id
+        # MD5 is used purely for deterministic compact hashing, not security.
+        return "call_" + hashlib.md5(raw_id.encode("utf-8")).hexdigest()
+
+    @staticmethod
     def _normalize_tool_call_ids(payloads: dict) -> None:
-        """Normalize oversized tool_call IDs before sending the request.
+        """Normalize oversized tool_call IDs in outgoing payloads.
 
-        Some OpenAI-compatible relay services return tool_call IDs that far
-        exceed the 64-character limit enforced by the OpenAI API spec
+        Some OpenAI-compatible relay services return tool_call IDs that
+        far exceed the 64-character limit enforced by the OpenAI API spec
         (observed lengths of 660 / 1650+ chars in the wild). Round-tripping
-        those IDs into the next request's messages triggers HTTP 400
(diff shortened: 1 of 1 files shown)


x1051445024:
Thanks for the review. Addressed in the follow-up commit:

1. **Indentation / `@staticmethod` decorator** — verified runtime-correct
   locally (AstrBot boots and the warning log fires on real traffic). The
   apparent misalignment was an artifact of the diff view with leading
   `+` markers.

2. **Duplicated shortening logic** — extracted into
   `_shorten_tool_call_id`, now shared by `_normalize_tool_call_ids` and
   `_parse_openai_completion`. The helper is documented as explicitly
   non-cryptographic.

3. **Log possibly leaking raw IDs** — log now only includes the original
   ID length and the normalized short form, not any prefix of the raw ID.

PR #8157, opened 2026-05-12
feat: add openai responses provider

mapleluvr:
## Summary
- Add an OpenAI Responses provider type with template exposure, runtime request mapping, response parsing, streaming support, and provider path tests.
- Preserve the existing OpenAI chat-completions provider path with regression coverage.
- Include the current root master checkpoint/subagent runtime updates already present locally before this merge.

## Test Plan
- `uv run pytest tests/test_openai_source.py -q -k "openai_responses or responses"`
- `uv run pytest tests/test_dashboard.py::test_provider_templates_include_openai_responses_type -q`
- `node --test tests/providerUtils.test.mjs` from `dashboard/`
- `uv run ruff check astrbot/core/provider/manager.py astrbot/core/provider/sources/openai_source.py tests/test_openai_source.py`


Soulter:
too many changes, closing...

PR #8185, opened 2026-05-14
fix: Faiss read/write failure on Windows with non-ASCII user paths

irmia2026:
## Summary

Fix Faiss index read/write failure on Windows when the path contains non-ASCII characters (Chinese, Japanese, Korean, Cyrillic, etc.).

## Root Cause

Faiss C++ (`io.cpp`) uses `fopen(const char*)` which expects ANSI codepage on Windows, but Python passes paths as UTF-8 bytes through the C extension bridge. Multi-byte characters get garbled → `No such file or directory`.

## Impact

All Windows users with non-ASCII usernames are unable to use the knowledge base feature. Upload fails with:
> 存储失败：文本块已生成，但写入知识库索引时出错

While chunks are written to `doc.db` (SQLite via Python), the Faiss vector index and `kb_documents` metadata are both lost.

## Fix

Temp-file bridge: Faiss reads/writes a pure-ASCII temp file; Python `shutil` (using Windows wide-char APIs natively) handles the real path.

Additional safeguards:
- `_safe_temp_dir()` — priority cascade ensuring ASCII temp directory even when `TEMP` env points to a non-ASCII path (falls back to `%SystemRoot%\Temp`)
- PID-based temp filenames for concurrency safety
- `os.makedirs` before `shutil.move` for first-time writes in nested directories

## Files Changed

One file: `app/astrbot/core/db/vec_db/faiss_impl/embedding_storage.py`
- New: `_safe_temp_dir()`, `_read_index()`, `_write_index()`
- Modified: `__init__()`, `save_index()`
- No public API changes. Fully backward-compatible — on Linux/macOS/ASCII-Windows, the `try:` direct read succeeds and the fallback is never triggered.

## Verification

7 tests passed on Windows with Chinese username path: write/read round-trip, concurrent access, first-time write, nested directories, polluted TEMP fallback.

## Summary by Sourcery

Add a Faiss embedding storage wrapper that safely reads and writes indices, including on Windows paths with non-ASCII characters.

Bug Fixes:
- Resolve Faiss index read/write failures on Windows when the target path contains non-ASCII characters by routing I/O through ASCII-only temporary files.

Enhancements:
- Introduce a robust EmbeddingStorage abstraction around Faiss indices with async insert, batch insert, search, delete, and persistence operations.
- Add a safe temporary-directory selection strategy to ensure ASCII-only, writable temp locations are used for Faiss I/O bridging on Windows.

### DIFF
--- a/astrbot/core/db/vec_db/faiss_impl/embedding_storage.py
+++ b/astrbot/core/db/vec_db/faiss_impl/embedding_storage.py
@@ -5,8 +5,69 @@ except ModuleNotFoundError:
         "faiss 未安装。请使用 'pip install faiss-cpu' 或 'pip install faiss-gpu' 安装。",
     )
 import os
+import shutil
+import tempfile
+import uuid
 
 import numpy as np
+
+
+# ── Faiss C++ fopen() 在 Windows 上使用 ANSI codepage ──
+# Python 传给 Faiss 的路径是 UTF-8 字节，但 Windows fopen 期望 ANSI 编码，
+# 导致含非 ASCII 字符的路径（如 C:\Users\中文用户名\...）被解读为乱码而失败。
+# 本模块通过"纯 ASCII 临时文件桥接"规避此问题。
+#
+# tempfile.gettempdir() 可能返回含中文用户的路径（取决于 TEMP 环境变量），
+# 所以 _safe_temp_dir() 硬编码一个保证纯 ASCII 且可写的目录。
+
+
+def _safe_temp_dir() -> str:
+    """返回保证纯 ASCII 且可写的临时目录，用于 Faiss I/O 桥接。
+
+    优先级:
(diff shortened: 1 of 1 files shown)


irmia2026:
All issues raised by gemini-code-assist and sourcery-ai have been addressed in f9a13d8. Ready for human review. 🙏

Soulter:
Your code path is wrong..

embedding_storage.py is at `astrbot/core/db/vec_db/faiss_impl/embedding_storage.py`

irmia2026:
Ah, good catch — that was a leftover `app/` prefix from how the file tree looked under my local AstrBot installation. I should have verified against the upstream layout before pushing.

Fixed in the latest force-push: the file is now at `astrbot/core/db/vec_db/faiss_impl/embedding_storage.py` where it belongs. Thanks for the review! 🙏

PR #8367, opened 2026-05-27
fix: emit final response after stream parse fallback

### LINKED ISSUE #8366: OpenAI-compatible streaming can finish without final assistant response
### Describe the bug

For OpenAI-compatible streaming providers, `_query_stream()` can yield content chunks but fail while building/parsing the final completion snapshot. The current code logs the final completion error and returns without yielding a non-chunk final `LLMResponse`.

Agent runners rely on the final non-chunk response to transition to DONE. If only streaming chunks are yielded, the user may see streamed text while the agent does not complete correctly and may continue until max steps or be marked complete by an outer runtime without a proper final response.

### Expected behavior

If stream chunks already produced assistant text/reasoning, the provider should still yield a final assistant `LLMResponse` from the streamed content so the runner can finish normally. If no usable streamed output exists, the original error should still surface.

### Reproduction idea

Mock an OpenAI streaming response that yields a text delta, then make final completion parsing raise `EmptyModelOutputError`. `_query_stream()` currently returns only the chunk and no final response.

### Affected area

`astrbot/core/provider/sources/openai_source.py`

zouyonghe:
## Summary

- preserve streamed text, reasoning, and usage while reading OpenAI-compatible streaming chunks
- emit a final non-chunk assistant response from streamed content if final completion parsing fails
- re-raise the final parsing error when no usable streamed output exists

Fixes #8366

## Tests

- `uv run pytest tests/test_openai_source.py`
- `uv run ruff check astrbot/core/provider/sources/openai_source.py tests/test_openai_source.py`

## Summary by Sourcery

Ensure OpenAI-compatible streaming queries emit a usable final assistant response when final completion parsing fails while preserving streamed output.

Bug Fixes:
- Return a synthesized final assistant response from accumulated streaming chunks when final completion parsing fails but partial output exists.
- Propagate final completion parsing errors when no usable streamed content was produced to avoid silent failures.

Enhancements:
- Capture and aggregate streamed text, reasoning content, and usage metadata during OpenAI-compatible streaming responses for reuse in the final response.

Tests:
- Add a streaming query test that verifies a final non-chunk assistant response is emitted when final completion parsing raises an empty model output error.

### DIFF
--- a/astrbot/core/provider/sources/openai_source.py
+++ b/astrbot/core/provider/sources/openai_source.py
@@ -655,6 +655,9 @@ class ProviderOpenAIOfficial(Provider):
         llm_response = LLMResponse("assistant", is_chunk=True)
 
         state = ChatCompletionStreamState()
+        streamed_text_parts: list[str] = []
+        streamed_reasoning_parts: list[str] = []
+        latest_usage = None
 
         async for chunk in stream:
             choice = chunk.choices[0] if chunk.choices else None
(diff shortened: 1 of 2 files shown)


zouyonghe:
Closing this PR because the root cause was misattributed. The observed failure is in the Multica runtime completion/comment-delivery contract, not in AstrBot's OpenAI provider.

PR #8547, opened 2026-06-03
fix(openai-embedding): enhance stability, empty query handling, and SiliconFlow domain compatibility

Rat0323:
This PR introduces several robustness and stability improvements to the `openai_text_embedding` provider to prevent crashes, handle empty queries gracefully, and correctly support both SiliconFlow domain configurations.

### Key Changes
1. **SiliconFlow Domain Check**: Refactored the dimensions removal logic to support both `.cn` and `.com` domains case-insensitively (using `urlparse` to safely extract the hostname) to prevent 400 Bad Request errors (`code 20015`).
2. **Fail-Fast Key Check**: Added initialization validation to verify that `embedding_api_key` is configured, failing fast with a descriptive error instead of throwing a generic error later.
3. **Safe Timeout Parsing**: Wrapped `timeout` parsing in a `try-except` block to fall back to `20` if the value is empty or invalid, avoiding `ValueError` crashes.
4. **Empty Query Protection**: Guarded single and batch retrieval entry points to return/raise errors early if input texts are empty or whitespace-only, preventing 400 Bad Request failures.
5. **Exception Handling**: Wrapped API exceptions in standard `openai.OpenAIError` blocks to print descriptive logs during network or API request failures.
6. **Resource Cleanup**: Protected client termination by checking `getattr(self, 'client', None)` inside `terminate()` to avoid attribute error exceptions during teardown.

### DIFF
--- a/astrbot/core/provider/sources/openai_embedding_source.py
+++ b/astrbot/core/provider/sources/openai_embedding_source.py
@@ -1,6 +1,7 @@
 from urllib.parse import urlparse
 
 import httpx
+import openai
 from openai import AsyncOpenAI
 
 from astrbot import logger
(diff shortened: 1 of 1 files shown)


Rat0323:
Closing this PR because its core embedding-dimensions and SiliconFlow handling has been superseded by #9245, which provides a more complete solution. Thanks.

PR #8588, opened 2026-06-05
fix(core): make StarTools.get_data_dir() robust under submodules / debug environments

Rat0323:
### Problem

`StarTools.get_data_dir()` can infer the plugin name when `plugin_name` is omitted, but the current implementation only checks `star_map` with the immediate caller module name.

That fails in common plugin development scenarios:

1. A plugin calls the API from a submodule, such as `data.plugins.example.services.cache`, while `star_map` only contains the registered entry module, such as `data.plugins.example.main`.
2. A plugin module is exercised through a local debug shim or `__main__` module where the module name itself is not present in `star_map`, even though the file still lives under the registered plugin directory.

In these cases, `get_data_dir()` raises at runtime even though the caller still belongs to a loaded plugin.

### Ecosystem Evidence

This is not limited to a single plugin. Several plugin-level PRs reference this core PR and work around the same failure by explicitly passing `plugin_name` to `StarTools.get_data_dir()`:

- SXP-Simon/astrbot_plugin_qq_group_daily_analysis#193 (open)
- kawayiYokami/astrbot_plugin_angel_heart#49 (merged)
- kawayiYokami/astrbot_plugin_angel_smile#7 (open)
- lxfight-s-Astrbot-Plugins/astrbot_plugin_livingmemory#169 (merged)
- olozhika/astrbot_plugin_local_reminiscence#31 (merged)
- YUMU1658/astrbot_plugin_qq_tools#5 (merged)
- Renyus/astrbot_plugin_self_evolution#123 (open)

The merged plugin-side PRs are useful backward-compatible fixes for existing AstrBot installations, and the open ones show the same pattern is still being handled in plugin repositories. This core PR remains useful because it fixes the framework helper itself, so future plugin submodules and debug entrypoints do not need to repeat the same workaround.

### Solution

Make plugin-name resolution more robust while keeping explicit `plugin_name` behavior unchanged:

1. If `plugin_name` is provided, use it directly.
2. If omitted, first try the existing exact `star_map` lookup.
3. If that fails, resolve the caller by registered plugin root, supporting plugin submodules under `data.plugins.<plugin>` and built-in plugin submodules under `astrbot.builtin_stars.<plugin>`.
4. If the caller module name is not useful, resolve its `__file__` relative to the configured plugin roots and map that root back to existing `StarMetadata`.
5. If no registered plugin can be identified, keep the existing hard failure instead of silently writing to a shared fallback directory.

This intentionally avoids a generic `unknown_plugin` fallback so unrelated or truly unknown modules do not share the same plugin data directory.

### Verification

- `uv run pytest tests/unit/test_star_tools.py -q --basetemp=.pytest-tmp-star-tools`
- `uv run ruff format --check .`
- `uv run ruff check .`

### Checklist

- [x] My changes have been tested, and verification steps are provided above.

### DIFF
--- a/astrbot/core/star/star_tools.py
+++ b/astrbot/core/star/star_tools.py
@@ -18,11 +18,11 @@
 """
 
 import inspect
-import logging
 import os
 import uuid
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Mapping
 from pathlib import Path
+from types import ModuleType
 from typing import Any, ClassVar
 
 from astrbot.api.platform import AstrBotMessage, MessageMember, MessageType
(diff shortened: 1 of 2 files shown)


Rat0323:
CI note: the current `Run pytest suite` failure does not appear to be caused by this PR.

The failing tests are:

- `tests/test_fastapi_v1_dashboard.py::test_v1_openapi_alias_websocket_routes_are_mounted`
- `tests/test_fastapi_v1_dashboard.py::test_dashboard_config_aliases_are_registered_on_fastapi`

I reproduced the same two failures on the current `origin/master` commit `40720fc2`, without the StarTools changes. The new `tests/unit/test_star_tools.py` cases added by this PR pass both locally and in the PR CI run.

PR #8742, opened 2026-06-12
fix(gemini): preserve assistant tool calls during history conversion

Rat0323:
## Summary

This PR fixes Gemini history conversion when an assistant message contains both normal content and `tool_calls`.

Previously, `_prepare_conversation()` handled assistant messages with mutually exclusive `if / elif / elif` branches. If an assistant message had text content, or list content such as a think/text part, the `tool_calls` branch was skipped. The next Gemini request could then contain a tool response without the corresponding assistant function-call part, which can make Gemini repeat the same tool call or request invalid tool names in later turns.

## Changes

- Preserve assistant text/list content and function calls in the same `ModelContent.parts` list.
- Treat `tool_calls: null` as empty, so OpenAI-style assistant messages with a null field do not crash local conversion.
- Add regression tests for:
  - assistant text content + `tool_calls`
  - assistant list/think content + `tool_calls`
  - `tool_calls: null`

## Notes

This PR was narrowed after re-checking the evidence. The earlier wording about a universal Gemini `thinking_config` + tools incompatibility was too broad, so that part has been removed from the patch and from the main claim here. This PR now focuses on the confirmed serialization bug.

## Verification

```bash
uv run python -m pytest tests/test_gemini_source.py -q
uv run ruff check astrbot/core/provider/sources/gemini_source.py tests/test_gemini_source.py
```

### DIFF
--- a/astrbot/core/provider/sources/gemini_source.py
+++ b/astrbot/core/provider/sources/gemini_source.py
@@ -290,6 +290,11 @@ class ProviderGoogleGenAI(Provider):
                 else:
                     thinking_config.thinking_level = level
 
+        # If any tools (native or custom function declarations) are active for this request,
+        # we must disable thinking because Gemini API does not support them simultaneously.
+        if tool_list:
+            thinking_config = None
+
         return types.GenerateContentConfig(
             system_instruction=system_instruction,
             temperature=temperature,
--- a/astrbot/core/provider/sources/gemini_source.py
+++ b/astrbot/core/provider/sources/gemini_source.py
@@ -386,11 +386,10 @@ class ProviderGoogleGenAI(Provider):
                 append_or_extend(gemini_contents, parts, types.UserContent)
 
             elif role == "assistant":
-                if isinstance(content, str):
-                    parts = [types.Part.from_text(text=content)]
-                    append_or_extend(gemini_contents, parts, types.ModelContent)
+                parts = []
+                if isinstance(content, str) and content:
+                    parts.append(types.Part.from_text(text=content))
                 elif isinstance(content, list):
-                    parts = []
                     thinking_signature = None
                     text = ""
                     for part in content:
@@ -415,10 +414,8 @@ class ProviderGoogleGenAI(Provider):
                             thought_signature=thinking_signature,
                         )
                     )
-                    append_or_extend(gemini_contents, parts, types.ModelContent)
 
-                elif not native_tool_enabled and "tool_calls" in message:
-                    parts = []
+                if not native_tool_enabled and "tool_calls" in message:
                     for tool in message["tool_calls"]:
                         part = types.Part.from_function_call(
                             name=tool["function"]["name"],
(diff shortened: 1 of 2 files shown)


Soulter:
是否有相关文档链接，以及报错日志？》

Rat0323:
**Superseded by the latest revision.**

This earlier comment reflected the initial hypothesis that explicit `thinking_config` and tools were universally incompatible. After re-checking the evidence and running a minimal reproduction, I narrowed this PR and removed that claim from the patch.

The current confirmed fix is focused on Gemini history conversion: assistant messages that contain both `content` and `tool_calls` must serialize both the text/list content and the function-call parts. See the updated PR description and the latest comment for the current scope.

Rat0323:
**Updated validation note.**

The useful part of the earlier log review is the pre-patch tool-loop evidence: in a PR-summary request, Gemini returned assistant content together with a tool call, then later turns repeated `get_pull_request_files` and eventually requested invalid names such as `github.get_pull_request_files` / `github_get_pull_request_files`.

After narrowing the PR, I no longer use these logs as evidence for a universal `thinking_config + tools` incompatibility. The confirmed root cause covered by this PR is the old `if / elif / elif` history conversion dropping `tool_calls` whenever assistant `content` was also present.

The latest revision adds focused regression tests for assistant text/list content with `tool_calls`, plus `tool_calls: null` compatibility.

Rat0323:
**Superseded by the latest revision.**

This note referred to the earlier broader patch. The current PR no longer claims a universal Gemini thinking/tools conflict and is now verified with focused unit tests for Gemini history conversion.

Rat0323:
Updated this PR after re-checking the evidence and running a minimal reproduction against the PR base.

The confirmed issue is Gemini history conversion: when an assistant message contains both `content` and `tool_calls`, the previous mutually exclusive branches serialized only the content and dropped the function-call parts. Later turns could then include tool responses without the matching assistant function calls, which explains the repeated/invalid tool-call behavior observed in the logs.

I also re-tested explicit `thinking_config + tools` in my current Gemini-compatible environment and could not reproduce a universal server-side incompatibility, so I removed that part from this patch and no longer present it as the main root cause.

This revision now focuses on the verified serialization bug and adds regression tests for assistant text/list content with `tool_calls`, plus `tool_calls: null` compatibility.

Verification:

```bash
uv run python -m pytest tests/test_gemini_source.py -q
uv run ruff check astrbot/core/provider/sources/gemini_source.py tests/test_gemini_source.py
```

Rat0323:
CI note for the latest revision:

The current failing `Unit Tests / Run pytest suite` check is not from the Gemini changes in this PR. In the failed run, `tests/test_gemini_source.py` passed, and the two failures are both in `tests/test_fastapi_v1_dashboard.py`:

- `test_v1_openapi_alias_websocket_routes_are_mounted`
- `test_dashboard_config_aliases_are_registered_on_fastapi`

This PR currently only changes:

- `astrbot/core/provider/sources/gemini_source.py`
- `tests/test_gemini_source.py`

After comparing the latest green master run and this failed PR run, the important difference appears to be dependency resolution timing. The green master CI run installed `fastapi==0.136.3`, while this PR run installed `fastapi==0.137.0` because the repository does not commit a lockfile for CI. With FastAPI 0.137.0, included routers can appear as `_IncludedRouter` wrappers in `app.router.routes`; these tests currently inspect top-level route objects directly and fail because those wrappers do not expose `.path` and websocket paths are no longer visible through that old enumeration method.

So I am keeping this PR focused on the Gemini history-conversion regression instead of adding unrelated dashboard/test fixes here. I also tried to rerun the failed check, but GitHub requires repository admin permission for that action.

Rat0323:
Closing this in favor of #8787. The original hypothesis in this PR was narrowed after validation, so I opened a clean replacement PR with a single focused commit for the confirmed Gemini history-conversion bug.

PR #8782, opened 2026-06-14
fix: deduplicate send_message_t fix: deduplicate send_message_to_user calls to prevent duplicate messages

Rrttttttt:
问题
某些 LLM（已知触发模型：mimo）在同一响应中同时返回 `completion_text` 和 `send_message_to_user`
工具调用，或在连续多次响应中重复调用 `send_message_to_user`，导致用户收到重复消息。
复现
1. 使用 mimo 模型
2. 在群聊中 @机器人
3. 机器人偶尔会发送两条完全相同的消息
修复方案
1. `tool_loop_agent_runner.py`（根因修复）
当 LLM 响应中包含 `send_message_to_user` 工具调用时，抑制 `completion_text` / `result_chain` 的 yield，防止
RespondStage 重复发送相同内容。
2. `message_tools.py`（指纹去重）
在 `send_message_to_user` 工具内部，计算消息指纹（md5），30 秒内相同内容直接跳过，防止跨响应的重复调用。
测试
日志中看到 `检测到 send_message_to_user 工具调用，抑制 completion_text` 说明根因修复生效
日志中看到 `[send_message_to_user] 检测到重复发送，已跳过` 说明指纹去重生效

## Summary by Sourcery

Prevent duplicate user-visible messages when LLMs return both text and send_message_to_user tool calls or repeatedly invoke the tool across responses.

Bug Fixes:
- Suppress completion_text/result_chain output when a send_message_to_user tool call is present in the LLM response to avoid double-sending the same content.
- Deduplicate send_message_to_user calls within a short time window using a message fingerprint to skip repeated sends across responses.

Enhancements:
- Add logging around suppression and deduplication paths to aid in diagnosing duplicate message behavior.

### DIFF
--- a/astrbot/core/agent/runners/tool_loop_agent_runner.py
+++ b/astrbot/core/agent/runners/tool_loop_agent_runner.py
@@ -792,6 +792,13 @@ class ToolLoopAgentRunner(BaseAgentRunner[TContext]):
             await self._complete_with_assistant_response(llm_resp)
 
         # 返回 LLM 结果
+        # 当 LLM 同时返回 completion_text 和 send_message_to_user 工具调用时，
+        # 抑制 completion_text 的 yield，避免 respond 阶段重复发送相同内容。
+        # 这是 mimo 模型的已知问题：它会在同一响应中既输出文本又调用 send_message_to_user。
+        _has_send_message_tool = (
+            llm_resp.tools_call_name
+            and "send_message_to_user" in llm_resp.tools_call_name
+        )
         if llm_resp.reasoning_content:
             yield AgentResponse(
                 type="llm_result",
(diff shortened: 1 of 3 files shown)


zouyonghe (changes requested):
Blocking concern: this suppresses  /  whenever the model returns any  tool call.  can target a different session (admins are allowed to do that), so a model response that sends a notification elsewhere and also explains the action to the current user would now silently drop the current-user response. The same problem can happen when the tool payload differs from ; this is not necessarily a duplicate.

Please narrow the suppression to proven duplicates only, for example by comparing the tool target/message payload against the current response, or keep the runner behavior unchanged and rely on a scoped duplicate guard inside . The dedup guard itself should also be scoped to an agent run/tool-call context if possible, because a process-global 30s fingerprint can skip legitimate repeated sends of identical content.

zouyonghe (changes requested):
Blocking concern: this suppresses `completion_text` / `result_chain` whenever the model returns any `send_message_to_user` tool call. `send_message_to_user` can target a different session (admins are allowed to do that), so a model response that sends a notification elsewhere and also explains the action to the current user would now silently drop the current-user response. The same problem can happen when the tool payload differs from `completion_text`; this is not necessarily a duplicate.

Please narrow the suppression to proven duplicates only, for example by comparing the tool target/message payload against the current response, or keep the runner behavior unchanged and rely on a scoped duplicate guard inside `send_message_to_user`. The dedup guard itself should also be scoped to an agent run/tool-call context if possible, because a process-global 30s fingerprint can skip legitimate repeated sends of identical content.

PR #8783, opened 2026-06-14
test: stabilize route assertions for FastAPI 0.137

he-yufeng:
The current dependency range allows FastAPI 0.137.0 / Starlette 1.3.1 in CI. With that stack, these two route-registration tests can fail because they inspect `app.router.routes` internals directly: FastAPI now keeps included routers behind `_IncludedRouter`, which has no `path` attribute and can hide routes from the flat list.

The routes themselves are still registered and resolvable through FastAPI's public `url_path_for()` API, so this keeps the tests focused on the intended behavior: the expected aliases are mounted at the expected paths.

### Modifications / 改动点

- Assert the route aliases through `app.url_path_for()` instead of scanning route object class names / `path` attributes directly.
- Keep the change limited to `tests/test_fastapi_v1_dashboard.py`.

- [x] This is NOT a breaking change. / 这不是一个破坏性变更。

### Screenshots or Test Results / 运行截图或测试结果

Verification run locally:

```text
python -m py_compile tests\test_fastapi_v1_dashboard.py
python -m ruff check tests\test_fastapi_v1_dashboard.py
python -m ruff format --check tests\test_fastapi_v1_dashboard.py
python -m pytest tests\test_fastapi_v1_dashboard.py -q -p no:cacheprovider
uv run python -c "import fastapi, starlette, pytest; print('fastapi', fastapi.__version__); print('starlette', starlette.__version__); print('pytest', pytest.__version__)"
uv run python -m pytest tests\test_fastapi_v1_dashboard.py -q -p no:cacheprovider
```

The `uv run` environment resolved the same relevant versions seen in CI:

```text
fastapi 0.137.0
starlette 1.3.1
pytest 9.1.0
```

Both pytest runs passed:

```text
52 passed
```

---

### Checklist / 检查清单

- [x] 😊 If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc. / No new feature; test compatibility only.
- [x] 👀 My changes have been well-tested, **and "Verification Steps" and "Screenshots" have been provided above**.
- [x] 🤓 I have ensured that no new dependencies are introduced, OR if new dependencies are introduced, they have been added to the appropriate locations in `requirements.txt` and `pyproject.toml`.
- [x] 😮 My changes do not introduce malicious code.

## Summary by Sourcery

Stabilize FastAPI v1 dashboard tests against internal routing changes by asserting routes via FastAPI’s public URL resolution API.

Tests:
[...]

PR #8787, opened 2026-06-14
fix(gemini): preserve assistant tool calls during history conversion

Rat0323:
## Summary

This PR fixes Gemini conversation history conversion for assistant messages that contain both `content` and `tool_calls`.

Previously, `ProviderGoogleGenAI._prepare_conversation` handled assistant message fields through mutually exclusive branches. When an assistant message had normal content and tool calls at the same time, only the content was serialized and the function-call parts were dropped. Later turns could then contain tool responses without the matching assistant function calls in history.

The updated conversion serializes assistant content and tool calls independently, while also tolerating `tool_calls: null`.

## Changes

- Preserve assistant `tool_calls` even when assistant `content` is also present.
- Preserve list-style assistant content, including think/text parts, alongside tool calls.
- Treat `tool_calls: null` the same as no tool calls.
- Add focused regression tests for these Gemini history-conversion cases.

## Validation

```bash
uv run python -m pytest tests/test_gemini_source.py -q
uv run ruff check astrbot/core/provider/sources/gemini_source.py tests/test_gemini_source.py
```

Both commands pass locally.

## CI Note

The current full `Run pytest suite` CI failure is unrelated to this PR's Gemini diff. In the PR run, `tests/test_gemini_source.py` passed, and the only failures were two existing dashboard route tests in `tests/test_fastapi_v1_dashboard.py`:

- `test_v1_openapi_alias_websocket_routes_are_mounted`
- `test_dashboard_config_aliases_are_registered_on_fastapi`

The important difference appears to be dependency resolution timing. The earlier green master CI run installed `fastapi==0.136.3`, while this PR run installed `fastapi==0.137.0` because the repository does not commit a lockfile for CI. FastAPI 0.137.0 exposes included routers differently through `_IncludedRouter`, which affects tests that inspect `app.router.routes` directly.

### DIFF
--- a/astrbot/core/provider/sources/gemini_source.py
+++ b/astrbot/core/provider/sources/gemini_source.py
@@ -377,11 +377,10 @@ class ProviderGoogleGenAI(Provider):
                 append_or_extend(gemini_contents, parts, types.UserContent)
 
             elif role == "assistant":
-                if isinstance(content, str):
-                    parts = [types.Part.from_text(text=content)]
-                    append_or_extend(gemini_contents, parts, types.ModelContent)
+                parts = []
+                if isinstance(content, str) and content:
+                    parts.append(types.Part.from_text(text=content))
                 elif isinstance(content, list):
-                    parts = []
                     thinking_signature = None
                     text = ""
                     for part in content:
@@ -389,7 +388,7 @@ class ProviderGoogleGenAI(Provider):
                         if part.get("type") == "think":
                             thinking_signature = part.get("encrypted") or None
                         else:
-                            text += str(part.get("text"))
+                            text += str(part.get("text") or "")
 
                     if thinking_signature and isinstance(thinking_signature, str):
                         try:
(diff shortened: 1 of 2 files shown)


zouyonghe (approved):
Reviewed. The conversion now independently preserves assistant content and , including list-style think/text content and . The focused Gemini tests cover the regression shape. The current full pytest failure is unrelated and matches the FastAPI route-test compatibility issue fixed by #8783.

zouyonghe (approved):
Reviewed. The conversion now independently preserves assistant content and tool_calls, including list-style think/text content and tool_calls null. The focused Gemini tests cover the regression shape. The current full pytest failure is unrelated and matches the FastAPI route-test compatibility issue fixed by #8783.

Rat0323:
Closing this PR because the core assistant tool-call conversion fix has landed through #8833. Any remaining tool_calls=null hardening can be proposed separately as a smaller follow-up. Thanks for the review.

PR #8790, opened 2026-06-15
fix: preserve guarded run-based tools

### LINKED ISSUE #8788: [bug] permission-wrapped FunctionTool run() tools return no callable handler
### What happened

A plugin tool registered via `Context.add_llm_tools()` can be exposed to the model successfully, but calling it returns:

```text
error: tool has no callable handler
```

This happens for dataclass-style `FunctionTool` subclasses that implement `run()` instead of `call()` or `handler`.

Example shape:

```python
from dataclasses import dataclass, field

from astrbot.api import FunctionTool
from astrbot.api.event import AstrMessageEvent


@dataclass
class ListMCPServersTool(FunctionTool):
    name: str = "list_mcp_servers"
    description: str = "List MCP servers."
    parameters: dict = field(default_factory=lambda: {"type": "object", "properties": {}})

    async def run(self, event: AstrMessageEvent):
        return "ok"
```

The tool appears in the available function-calling tools and the model can call it, but execution returns `error: tool has no callable handler`.

### Log evidence

```text
Agent 使用工具: ['list_mcp_servers']
使用工具：list_mcp_servers，参数：{}
Tool `list_mcp_servers` Result: error: tool has no callable handler
```

### Expected behavior

Dataclass-style `FunctionTool` subclasses that implement `run()` should continue to work when registered with `Context.add_llm_tools()`, including after the per-tool permission wrapper is applied.

### Suspected cause

`FunctionToolExecutor._execute_local()` supports tools with `handler`, overridden `call()`, or `run()`:

```python
if not tool.handler and not hasattr(tool, "run") and not is_override_call:
    raise ValueError("Tool must have a valid handler or override 'run' method.")
...
elif hasattr(tool, "run"):
    awaitable = getattr(tool, "run")
    method_name = "run"
```

However, `FunctionToolManager.get_full_tool_set()` wraps non-builtin tools in `_PermissionGuardedTool`, and `_PermissionGuardedTool.call()` currently only delegates to:

- `self._wrapped.handler`
- `self._wrapped.call()` if `call()` is overridden

It does not delegate to `self._wrapped.run()`.

As a result, any plugin tool that relies on `run()` becomes a wrapped tool with no recognized callable path and returns:

```python
return "error: tool has no callable handler"
```

### Why this is not just a plugin bug

The documented/dataclass-style plugin pattern supports `run()` implementations. A related historical issue (#8143) includes an example using `FunctionTool` with `async def run(...)`, and `FunctionToolExecutor._execute_local()` still supports this contract.

The incompatibility appears specifically after `_PermissionGuardedTool` was introduced for non-builtin tools.

### Possible fix

Update `_PermissionGuardedTool.call()` to mirror `_execute_local()` delegation behavior and support `run()` as a fallback after `handler` and overridden `call()`.

Pseudo-fix:

```python
run = getattr(self._wrapped, "run", None)
if run is not None:
    result = run(context.context.event, **kwargs)
    if inspect.isasyncgen(result):
        last = None
        async for item in result:
            last = item
        return last
    if inspect.isawaitable(result):
        return await result
    return result
```

The exact event/context argument handling should match the existing `call_local_llm_tool(..., method_name="run")` behavior.

### Environment

- AstrBot version observed in logs: `v4.25.5`
- Tool registration path: third-party plugin using `Context.add_llm_tools(ListMCPServersTool(...))`
- Tool implementation style: dataclass subclass of `FunctionTool` with `async def run(...)`

zouyonghe:
## Summary
- delegate permission-guarded FunctionTool wrappers to underlying run() methods when no handler or overridden call() exists
- keep run() argument semantics aligned with FunctionToolExecutor by passing the event object
- add a regression test for dataclass-style plugin tools that implement run()

Fixes #8788.

## Verification
- uv run pytest tests/unit/test_tool_permission.py -q
- uv run ruff check astrbot/core/provider/func_tool_manager.py tests/unit/test_tool_permission.py
- uv run ruff format --check astrbot/core/provider/func_tool_manager.py tests/unit/test_tool_permission.py
- git diff --check

## Summary by Sourcery

Ensure permission-guarded tools correctly delegate to underlying run-based tools when no custom call handler is defined.

Bug Fixes:
- Fix permission-guarded tools so they invoke the wrapped tool's run() method when no handler or overridden call() exists, returning the final result from sync, async, or async-generator implementations.

Tests:
- Add a regression test verifying that permission-guarded tools delegate to a wrapped tool's run() method and pass through the event context.

### DIFF
--- a/astrbot/core/provider/func_tool_manager.py
+++ b/astrbot/core/provider/func_tool_manager.py
@@ -266,6 +266,19 @@ class _PermissionGuardedTool(FunctionTool):
         if call_override is not None and call_override is not FunctionTool.call:
             return await self._wrapped.call(context, **kwargs)
 
+        run = getattr(self._wrapped, "run", None)
+        if run is not None:
+            event = context.context.event
+            result = run(event, **kwargs)
+            if _inspect.isasyncgen(result):
+                last: Any = None
+                async for item in result:
+                    last = item
+                return last
+            if _inspect.isawaitable(result):
+                return await result
+            return result
+
         return "error: tool has no callable handler"
 
 
--- a/tests/unit/test_tool_permission.py
+++ b/tests/unit/test_tool_permission.py
@@ -255,6 +255,27 @@ async def test_guarded_tool_delegates_to_wrapped_call():
 
 
 @pytest.mark.asyncio
+async def test_guarded_tool_delegates_to_wrapped_run():
+    _clear_tool_permissions()
+    mgr = FunctionToolManager()
+
+    class RunnableTool(FunctionTool):
+        async def run(self, event, **kwargs):
+            return f"from run(): {event.get_sender_id()} {kwargs['value']}"
+
+    wrapped = RunnableTool(
+        name="has_run",
+        description="desc",
+        parameters={},
+    )
+    guarded = _PermissionGuardedTool(wrapped, mgr)
+    context = _make_context(sender_id="runner")
+
+    result = await guarded.call(context, value="ok")
+    assert result == "from run(): runner ok"
+
+
+@pytest.mark.asyncio
 async def test_guarded_tool_handles_async_generator_handler():
     _clear_tool_permissions()
     mgr = FunctionToolManager()


zouyonghe:
CI note: the full pytest job failure is unrelated to this PR. It fails only in tests/test_fastapi_v1_dashboard.py under FastAPI 0.137 route internals, matching the compatibility issue fixed by #8783. The targeted tool permission tests and all other checks for this PR pass.

PR #8820, opened 2026-06-16
fix(computer): enforce shell timeout on Windows by killing the process tree

he-yufeng:
On Windows, the local computer-use shell tool ignores its `timeout`. A command given `timeout=2` that spawns a child sleeping for 30s still blocks for ~30s before returning.

Root cause is in `LocalShellComponent.exec`. It runs the command via `subprocess.run(..., shell=True, capture_output=True, timeout=...)`. With `shell=True` the command runs as a child of `cmd.exe`. When the timeout fires, `subprocess.run` calls `Popen.kill()`, which terminates only `cmd.exe`. The real worker keeps running and keeps the captured stdout/stderr pipes open, so the follow-up `communicate()` blocks until that child finally exits. The timeout is effectively a no-op, and since `exec` runs in a worker thread (`asyncio.to_thread`), the thread stays tied up the whole time.

This does not bite on POSIX, where the shell typically `exec`s a single command in place so there is no surviving child to leak the pipe.

### Modifications / 改动点

`astrbot/core/computer/booters/local.py`:
- Non-background branch now uses `subprocess.Popen(...) as proc` + `proc.communicate(timeout=...)`.
- On `TimeoutExpired`, kill the whole process tree, then drain/reap and re-raise. New helper `_kill_process_tree` runs `taskkill /F /T /PID <pid>` on Windows (tears down the tree) and falls back to `proc.kill()` on POSIX, so POSIX behavior is unchanged. The drain mirrors CPython's own `subprocess.run` (`communicate()` on Windows, `wait()` on POSIX).
- The same `TimeoutExpired` is re-raised, so the upper-layer error string is unchanged.

### Verification Steps / Test Results

Reproduced on current `master` (Windows 11, Python 3.13) with a real subprocess: a 30s sleep under `timeout=2` returned after 30.1s.

Added `tests/test_local_shell_component.py::test_local_shell_component_timeout_returns_promptly` (sleeps 30s, `timeout=2`, asserts it returns in <15s):
- Before the fix: `FAILED ... AssertionError: shell timeout was not enforced (took 30.1s)`
- After the fix: same command returns in ~2.5s with the identical `Command '...' timed out after 2 seconds` message.

The 4 existing decode tests mocked `subprocess.run`; since the code now uses `Popen`, they were switched to a small `_FakePopen` context manager exposing `communicate()`/`returncode`.

```
$ python -m pytest tests/test_local_shell_component.py -q
5 passed in ~6.8s
```

`ruff format --check` and `ruff check` pass on the changed source file.

### Checklist / 检查清单

- [x] My changes have been well-tested, and verification steps/results are provided above.

## Summary by Sourcery

Ensure local shell commands respect execution timeouts by correctly handling process termination across platforms.

Bug Fixes:
- Fix Windows shell executions ignoring timeouts by killing the entire process tree when a timeout occurs.

Tests:
- Add an integration-style timeout test for LocalShellComponent to verify that long-running shell commands return promptly on timeout.
- Update LocalShellComponent decoding tests to use a fake Popen context manager instead of mocking subprocess.run.

### DIFF
--- a/astrbot/core/computer/booters/local.py
+++ b/astrbot/core/computer/booters/local.py
@@ -44,6 +44,30 @@ def _is_safe_command(command: str) -> bool:
     return not any(pat in cmd for pat in _BLOCKED_COMMAND_PATTERNS)
 
 
+def _kill_process_tree(proc: subprocess.Popen) -> None:
+    """Forcefully terminate a process and every child it spawned.
+
+    On Windows ``Popen.kill()`` only terminates the immediate (shell) process.
+    Any child it started keeps running and holds the captured stdout/stderr
+    pipes open, so ``communicate()`` blocks until those children exit and the
+    requested timeout is never actually enforced. ``taskkill /T`` tears down the
+    whole tree. On POSIX we keep the existing ``proc.kill()`` behavior.
+    """
+    if os.name == "nt":
+        try:
+            subprocess.run(
+                ["taskkill", "/F", "/T", "/PID", str(proc.pid)],
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
+                check=False,
+            )
+            return
+        except OSError:
+            # taskkill missing/unavailable: fall back to a single-process kill.
+            pass
+    proc.kill()
+
+
 def _decode_bytes_with_fallback(
     output: bytes | None,
     *,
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this one, master already covers it. The timeout path in `local.py` now runs `taskkill /F /T /PID` on Windows (falling back to `proc.kill()`), which tears down the whole process tree, so this change is no longer needed. Thanks!

PR #8992, opened 2026-06-24
fix: refine context compression fallback

Soulter:
## Summary
- reuse split_into_rounds() for max-turn counting and turn-based truncation so tool/function messages stay in their logical round
- keep LLM compression as the first choice for llm_compress, then hard truncate repeatedly when compression fails or still exceeds the token threshold
- run request context compaction on every tool-loop step instead of limiting it to the first step, while keeping hard-truncated contexts out of persisted history

This is a smaller replacement/follow-up for #8974.

## Tests
- uv run ruff format .
- uv run ruff check .
- uv run pytest tests/agent/test_context_manager.py tests/agent/test_truncator.py

## Summary by Sourcery

Refine context truncation and compression behavior to treat tool call chains as single rounds, prefer LLM-based compression with robust fallbacks, and keep hard-truncated contexts out of persisted history while updating tests accordingly.

New Features:
- Expose context processing metadata via a new process_with_meta API that reports whether hard truncation occurred.
- Track and persist a separate, non-truncated message history alongside the working context in the agent run context.

Bug Fixes:
- Ensure turn-based truncation counts tool/function call chains as a single logical round so they are not over-truncated.
- Fix context compaction in the tool-loop agent runner so it runs on every step without polluting persisted history when hard truncation occurs.
- Handle LLM compression failures by falling back to deterministic hard truncation until the context fits within token limits.

Enhancements:
- Refine token-based compression to repeatedly drop oldest rounds before halving, avoiding regressions in token counts.
- Extend the LLM summary compressor to report failures explicitly so callers can distinguish lossy hard truncation from successful summarization.
- Use shared round-splitting utilities for both truncation and round counting to keep tooling messages grouped logically.

Tests:
- Add tests for round counting with tool chains, LLm compression failure fallbacks, and the updated truncation behavior.
- Introduce a deterministic token counter utility for stable, predictable compression tests.


Soulter:
Closing because this was based on #8974; will reopen a clean PR based directly on latest master.

PR #9081, opened 2026-06-30
fix: remove unauthenticated hosted core updates

LIghtJUNction:
### Motivation

- Close a high-risk update authenticity gap by removing the updater path that preferred a predictable hosted core package URL and accepted any syntactically valid ZIP without an authenticity binding.
- Ensure core updates remain tied to release metadata (`zipball_url`) or explicit GitHub archive URLs to prevent arbitrary code execution from a compromised CDN/storage path.

### Description

- Remove the hosted core package base URL and the `_build_core_package_url` logic so the updater no longer constructs or prefers `.../download/astrbot-core/{tag}/source.zip`.
- Remove the code path that downloaded the hosted package and only checked `zipfile.is_zipfile()` before accepting it; the updater now always downloads the release `zipball_url` (or the GitHub archive for commit hashes).
- Update unit tests in `tests/test_updator_socks.py` to assert that the release `zipball_url` is used for core package downloads and to remove assertions that relied on the hosted-package behavior.
- Keep extraction/apply behavior unchanged (the fix narrows the trusted update root rather than changing extract semantics). 

### Testing

- Ran `ruff format` and `ruff check` on the modified files (`astrbot/core/updator.py` and `tests/test_updator_socks.py`), which completed successfully locally.
- Attempted to run `pytest` for the targeted updater tests, but the test run was blocked by missing test runtime dependency `pytest_asyncio` in the environment, so `pytest` could not be completed.
- `uv`-based formatting/check attempts were blocked by network/cache dependency resolution in this environment, and `git diff --check` reported no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3f39592c1c8320ac3d5d500fea5459)

## Summary by Sourcery

Remove the unauthenticated hosted core update path so core updates are always fetched from trusted release metadata or GitHub archives.

Bug Fixes:
- Eliminate the ability to download and apply core update ZIPs from a predictable hosted URL without authenticity binding.

Tests:
- Update updater tests to assert core packages are downloaded from the release zipball URL only and remove hosted-package fallback scenarios.

### DIFF
--- a/astrbot/core/updator.py
+++ b/astrbot/core/updator.py
@@ -1,7 +1,6 @@
 import os
 import sys
 import time
-import zipfile
 from pathlib import Path
 
 import psutil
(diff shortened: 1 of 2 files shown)


LIghtJUNction:
这个还需要讨论

PR #9082, opened 2026-06-30
fix: disable workspace skills in group sessions

LIghtJUNction:
### Motivation

- Prevent a privilege-boundary vulnerability where non-admin users can plant `skills/<name>/SKILL.md` in a shared group workspace and have those workspace skills injected into later admin requests when `computer_use_runtime` is `local`.

### Description

- Restrict request-scoped workspace skill discovery in `astrbot/core/astr_main_agent.py` to only run when `runtime == "local"` and the event is not a group session by checking `not event.get_group_id()`.
- Add a regression test in `tests/unit/test_astr_main_agent.py` that creates a workspace skill in a group-session workspace and asserts it is not injected into the system prompt.
- Files changed: `astrbot/core/astr_main_agent.py`, `tests/unit/test_astr_main_agent.py`.

### Testing

- Ran `ruff format` and `ruff check` against the modified files and the checks passed.
- Added a unit test that verifies workspace skills are skipped for group sessions; running it via `pytest -k 'workspace_skills'` was attempted but blocked by dependency fetch/network tunnel errors in the environment.
- Attempts to run `uv run ruff format` and `uv run pytest` were also blocked by the same dependency fetch/network tunnel issue, so full CI-style execution could not complete in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a3f3b02c41083209108867997884e59)

## Summary by Sourcery

Guard workspace skill discovery against group sessions to avoid injecting group-scoped skills into admin/local requests.

Bug Fixes:
- Prevent workspace skills defined in shared group workspaces from being injected into system prompts when the runtime is local.

Tests:
- Add a unit test ensuring workspace skills are skipped for group sessions and not injected into the system prompt.


Soulter:
Would a more reasonable approach be to prohibit non-admiins from creating skills in the workspace?

LIghtJUNction:
> Would a more reasonable approach be to prohibit non-admiins from creating skills in the workspace?

make it configurable, but keep the default strict. That provides a better balance between flexibility and security.

PR #9090, opened 2026-06-30
fix(core): return clear error when file read tool receives a directory path

THE-FLIER:
## Summary
...

## Summary by Sourcery

Clarify error handling when the file read tool is given invalid paths, especially directories, and cover the behavior with tests.

Bug Fixes:
- Return a clear, user-friendly error when the file read tool is invoked with a directory path instead of a file.
- Normalize error propagation from the remote probe script so structured file read errors are surfaced instead of generic runtime failures.

Tests:
- Add an async test to verify the file read tool rejects directory paths with the new, descriptive error message and without leaking low-level OS error text.

### DIFF
--- a/astrbot/core/computer/file_read_utils.py
+++ b/astrbot/core/computer/file_read_utils.py
@@ -78,13 +78,26 @@ class ParsedDocument:
     text: str
 
 
+def _directory_read_error(path: str) -> str:
+    return (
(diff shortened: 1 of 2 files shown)


THE-FLIER:
Closing in favor of #9088, which fixes the same issue with a smaller, focused change. Thanks for the review feedback!

PR #9099, opened 2026-07-01
feat(provider): make outer OpenAI retry count configurable

x1051445024:
## Summary

- Add `provider_settings.provider_error_retries` to configure the OpenAI-compatible provider's outer recovery retry loop.
- Default the outer provider-level retry loop to 1 attempt to avoid multiplying latency with `request_max_retries`.
- Keep transport/status retry behavior controlled by the existing `request_max_retries` setting.
- Add unit coverage for coercing the new retry setting.

## Motivation

OpenAI-compatible proxy/aggregator providers often perform their own upstream retry and fallback. AstrBot currently has an inner request retry loop plus a fixed outer provider-level retry loop of 10 attempts. When the upstream proxy is slow or unstable, those nested retries can significantly delay fallback to other providers.

This change makes the outer loop configurable while preserving the existing recovery paths for users who need more attempts.

## Testing

- `python -m py_compile astrbot/core/provider/sources/openai_source.py astrbot/core/config/default.py tests/test_openai_source.py`

`pytest` was not available in the local environment used for this change.

## Summary by Sourcery

Make the OpenAI-compatible provider’s outer recovery retry loop configurable via provider settings and align defaults.

New Features:
- Add a provider_settings.provider_error_retries option to control provider-level recovery retries for the OpenAI-compatible provider.

Enhancements:
- Default provider-level recovery retries to 1 attempt and wire this setting into text and streaming chat flows to avoid compounding latency with request_max_retries.
- Extend default configuration schema to include the new provider_error_retries setting.

Tests:
- Add unit tests to verify defaulting and coercion behavior for the new provider_error_retries setting.

### DIFF
--- a/astrbot/core/config/default.py
+++ b/astrbot/core/config/default.py
@@ -104,6 +104,7 @@ DEFAULT_CONFIG = {
         "default_provider_id": "",
         "fallback_chat_models": [],
         "request_max_retries": 5,
+        "provider_error_retries": 1,
         "default_image_caption_provider_id": "",
         "image_caption_prompt": "Please describe the image using Chinese.",
         "provider_pool": ["*"],  # "*" 表示使用所有可用的提供者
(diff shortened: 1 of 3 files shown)


Soulter:
hi，感谢你的建议，不过目前不太打算将其可配置化，更应该说这个 retry 机制已经被新的 retry 机制替代了，之后可能会移除。

PR #9070, opened 2026-06-28
fix: Improve MCP client startup and shutdown handling to prevent idle spinning when a server is disabled and ensure proper cleanup on errors, timeouts, and cancellation

GlowingBrick:
#9068 

修复了
### Modifications / 改动点

修复了禁用mcp服务器时由于作用域不匹配出现的处理器空转问题
- [x] This is NOT a breaking change. / 这不是一个破坏性变更。

### Screenshots or Test Results / 运行截图或测试结果
禁用一个MCP，在Debug下：
```
[21:25:48.931] [Core] [INFO] [provider.func_tool_manager:720]: Connected to MCP server paddleocr, Tools: ['paddleocr_vl']
[21:25:52.805] [Core] [INFO] [provider.func_tool_manager:729]: Received shutdown signal for MCP client paddleocr
[21:25:53.138] [Core] [INFO] [provider.func_tool_manager:848]: Disconnected from MCP server paddleocr
```
可以注意到不再出现`Error closing current exit stack`错误

---

### Checklist / 检查清单

- [x] 😊 If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc. 
  / 如果 PR 中有新加入的功能，已经通过 Issue / 邮件等方式和作者讨论过。

- [x] 👀 My changes have been well-tested, **and "Verification Steps" and "Screenshots" have been provided above**.
  / 我的更改经过了良好的测试，**并已在上方提供了“验证步骤”和“运行截图”**。

- [x] 🤓 I have ensured that no new dependencies are introduced, OR if new dependencies are introduced, they have been added to the appropriate locations in `requirements.txt` and `pyproject.toml`.
  / 我确保没有引入新依赖库，或者引入了新依赖库的同时将其添加到 `requirements.txt` 和 `pyproject.toml` 文件相应位置。

- [x] 😮 My changes do not introduce malicious code.
  / 我的更改没有引入恶意代码。

## Summary by Sourcery

Improve MCP client startup and shutdown handling to prevent idle spinning when a server is disabled and ensure proper cleanup on errors, timeouts, and cancellation.

Bug Fixes:
- Prevent MCP client tasks from spinning when an MCP server is disabled due to mismatched cancellation scope.
- Ensure MCP client runtimes are correctly removed and cleaned up on connection errors, timeouts, or task cancellation.

Enhancements:
- Unify MCP client connection, tool registration, lifecycle management, and cleanup into a single task for safer resource handling.
[...]

PR #9132, opened 2026-07-04
fix: exit MCP anyio contexts in the lifecycle task on disable

kldsjfas:
#9068 is still reproducible on current master: disabling an MCP server exits the anyio cancel scopes from a task other than the one that entered them, which corrupts their state — the failure mode behind the reported 100% CPU spin.

#9070 introduced the per-server lifecycle task precisely so that cleanup happens in the task that entered the anyio contexts, but running the cleanup through `asyncio.shield()` defeats that: `shield()` wraps the coroutine in a *new* task, so `exit_stack.aclose()` still runs outside the owning task and fails with `Attempted to exit cancel scope in a different task than it was entered in` (visible at debug level) on every plain enable → disable, no cancellation involved.

### Modifications / 改动点

- `astrbot/core/provider/func_tool_manager.py`: await `_terminate_mcp_client()` directly in the lifecycle task's `finally` instead of wrapping it in `asyncio.shield()`. The graceful disable path has no cancellation in flight, and on forced shutdown the lifecycle task still runs its `finally` in-task, so the shield only served to break task affinity.
- `tests/unit/test_func_tool_manager.py`: regression test asserting that shutdown cleanup runs in the same task that connected (fails on master, passes with this fix).

Related: a reconnect from a tool call can still create an exit stack owned by a different task (#8056, addressed by the larger rework in #8307). This PR deliberately keeps to the disable path and does not touch `mcp_client.py`, so it does not conflict with #8307.

Enable → disable of a real stdio MCP server (FastMCP echo server), instrumented to record which task runs the cleanup:

On master (20008f1):

```text
cleanup in connect task: False
cancel scope errors: ['Error closing current exit stack: Attempted to exit cancel scope in a different task than it was entered in']
```

With this fix:

```text
cleanup in connect task: True
cancel scope errors: none
CPU after disable: cpu=0.00s over wall=3.00s (idle)
```

`uv run pytest tests/unit`: 666 passed; the 6 failures on this Windows machine (test_computer, test_message_tools, test_astr_main_agent) are identical on clean master. `ruff format --check` and `ruff check` pass on the changed files.

---

### Checklist / 检查清单

  / 我的更改没有引入恶意代码。

## Summary by Sourcery

Ensure MCP server disable cleanup runs in the same lifecycle task that established the connection to avoid anyio cancel scope corruption and CPU spin.

Bug Fixes:
- Fix MCP server disable path so exit stack and cleanup execute in the owning lifecycle task instead of a separate shielded task, preventing cross-task cancel scope errors and high CPU usage.

Tests:
- Add regression test verifying MCP shutdown cleanup runs in the same task that performed the connection and that the MCP client is removed after disable.

### DIFF
--- a/astrbot/core/provider/func_tool_manager.py
+++ b/astrbot/core/provider/func_tool_manager.py
@@ -723,8 +723,10 @@ class FunctionToolManager:
                 logger.debug(f"MCP client {name} task was cancelled")
                 raise
             finally:
-                # Cleanup in the same task that entered the anyio contexts
-                await asyncio.shield(self._terminate_mcp_client(name))
+                # Cleanup in the same task that entered the anyio contexts.
+                # No asyncio.shield() here: it would run the cleanup in a new
+                # task, and anyio cancel scopes cannot exit across tasks (#9068).
+                await self._terminate_mcp_client(name)
 
         lifecycle_task = asyncio.create_task(
             connect_and_lifecycle(), name=f"mcp-client:{name}"
--- a/astrbot/core/provider/func_tool_manager.py
+++ b/astrbot/core/provider/func_tool_manager.py
@@ -723,10 +723,19 @@ class FunctionToolManager:
                 logger.debug(f"MCP client {name} task was cancelled")
                 raise
             finally:
-                # Cleanup in the same task that entered the anyio contexts.
-                # No asyncio.shield() here: it would run the cleanup in a new
-                # task, and anyio cancel scopes cannot exit across tasks (#9068).
-                await self._terminate_mcp_client(name)
+                # Cleanup in the same task that entered the anyio contexts:
+                # asyncio.shield() would schedule the coroutine as a separate
+                # Task, and anyio cancel scopes cannot exit across tasks (#9068).
+                # Absorb late cancellations so a forced shutdown cannot abort
+                # the cleanup halfway.
+                task = asyncio.current_task()
+                while True:
+                    try:
+                        await self._terminate_mcp_client(name)
+                        break
+                    except asyncio.CancelledError:
+                        if task is not None:
+                            task.uncancel()
 
         lifecycle_task = asyncio.create_task(
             connect_and_lifecycle(), name=f"mcp-client:{name}"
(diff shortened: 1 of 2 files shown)


kldsjfas:
Addressed the cancellation concern in 419c448: if a cancellation lands while the finally-block cleanup is running, the cleanup is now retried in the same task with `Task.uncancel()`, so a forced shutdown can no longer abort it halfway (added a regression test for this too). Thanks for the `uncancel()` pointer.

On the `asyncio.shield()` mechanics: shield does run the wrapped coroutine in a separate task — per the CPython docs, *"If aw is a coroutine it is automatically scheduled as a Task"* (`shield()` calls `ensure_future()` internally). That task hop is exactly what made anyio raise `Attempted to exit cancel scope in a different task than it was entered in` on every disable: the regression test in this PR records `asyncio.current_task()` on both the connect and cleanup sides, and it fails on master because the two differ. I reworded the code comment to match the docs' phrasing ("schedule the coroutine as a separate Task").

Soulter:
@gemini-code-assist /gemini review

kldsjfas:
On the Python 3.10 compatibility concern: the project's supported floor is Python 3.12 — `pyproject.toml` declares `requires-python = ">=3.12"`, the Dockerfile base image is `FROM python:3.12-slim` (line 1, not 3.10-slim), and the README states "AstrBot requires Python 3.12 or later". `Task.uncancel()` has existed since Python 3.11, so it is available on every supported runtime; a `hasattr` guard would be unreachable code. Keeping the call unguarded.

Soulter:
> On the Python 3.10 compatibility concern: the project's supported floor is Python 3.12 — `pyproject.toml` declares `requires-python = ">=3.12"`, the Dockerfile base image is `FROM python:3.12-slim` (line 1, not 3.10-slim), and the README states "AstrBot requires Python 3.12 or later". `Task.uncancel()` has existed since Python 3.11, so it is available on every supported runtime; a `hasattr` guard would be unreachable code. Keeping the call unguarded.

Hi, many users are still using 3.10.x, it would be better if the change can compatible to 3.10

kldsjfas:
added the hasattr guard in 86094f2, so on 3.10 uncancel() is just skipped. should be fine there — catching the CancelledError is already enough for the retry, uncancel() is only 3.11+ bookkeeping. rest of the change doesn't use any 3.11+ API.

Soulter (approved):

Issue #9196, opened 2026-07-09
[Bug] Anthropic / Claude prompt caching hit rate is much lower than OpenAI under similar AstrBot usage


911218sky:
## What happened / 发生了什么

When I use Claude in AstrBot, prompt cache hits are much lower than expected.

Under a similar setup, GPT models can hit cache much more often, but Claude often creates cache again instead of reading from cache.

From my testing, this seems to be related to how AstrBot sends Anthropic requests, especially when tools/plugins are enabled.

## Reproduce / 如何复现？

1. Use a Claude model in AstrBot.
2. Enable tools or plugins.
3. Send several similar messages in the same conversation.
4. Check the provider-side usage stats.
5. Claude often does not hit cache well, while GPT performs much better in a similar setup.

## AstrBot version, deployment method (e.g., Windows Docker Desktop deployment), provider used, and messaging platform used. / AstrBot 版本、部署方式（如 Windows Docker Desktop 部署）、使用的提供商、使用的消息平台适配器

AstrBot v4.26.5  
Docker deployment  
Claude model via Anthropic-compatible provider path (`claude-sonnet-5`)  
Telegram adapter

## OS

Linux

## Logs / 报错日志

There may be no Python exception for this issue.

The main problem is visible in provider-side usage stats:
- Claude often creates cache again instead of reading cache.
- GPT cache hit rate is much better under similar usage.

I also tested a local patch and the behavior improved after:
- keeping tool order stable
- adding Anthropic cache-control markers on stable request parts

If needed, I can provide screenshots showing the before/after difference.

## Are you willing to submit a PR? / 你愿意提交 PR 吗？

Yes!

PR #9197, opened 2026-07-09
fix: use ephemeral cache control mode in anthropic provider to increase cache hit rate

911218sky:
Related to #9196

### Changes / 改动

- add explicit `cache_control` to the last `system` block in the Anthropic provider request payload

### Verification / 验证

- `python3 -m py_compile astrbot/core/provider/sources/anthropic_source.py`
- tested with the running `astrbot` Docker container after syncing the updated file
- confirmed the container loaded the updated code


Soulter (approved):
LGTM. Thanks for your contribution! ❤️

PR #9198, opened 2026-07-09
fix: update MiniMax TTS default model

octo-patch:
Reason: MiniMax TTS is integrated but still defaults to speech-02-turbo instead of the target speech-2.8-hd model.

## Summary
- Update the built-in MiniMax TTS provider template to default to `speech-2.8-hd`.

## Validation
- `uv run --group dev pytest tests/unit/test_config.py` (`41 passed`)
- `uv run --group dev ruff format --check astrbot/core/config/default.py tests/unit/test_config.py`
- `uv run --group dev ruff check astrbot/core/config/default.py tests/unit/test_config.py`
- `git diff --check`

### DIFF
--- a/tests/unit/test_config.py
+++ b/tests/unit/test_config.py
@@ -6,7 +6,7 @@ import os
 import pytest
 
 from astrbot.core.config.astrbot_config import AstrBotConfig, RateLimitStrategy
-from astrbot.core.config.default import CONFIG_METADATA_2, DEFAULT_VALUE_MAP
+from astrbot.core.config.default import DEFAULT_VALUE_MAP
 from astrbot.core.config.i18n_utils import ConfigMetadataI18n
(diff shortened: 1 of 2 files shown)


tests/unit/test_config.py
```diff
@@ -43,6 +43,15 @@ def minimal_default_config():
     }
 
 
+def test_minimax_tts_default_model_is_speech_28_hd():
```
Soulter: no need to add a new unit test for this change.

Soulter (changes requested):

octo-patch:
Thanks for the review. I removed the additional unit test as requested and pushed commit `465e6e49`. The final change is limited to the MiniMax TTS default configuration; I ran `uv run --group dev pytest tests/unit/test_config.py`, Ruff format and lint checks on the affected files, and `git diff --check`.

PR #9293, opened 2026-07-15
Fix GLM OpenAI-compatible content format handling

WINE92:
## Description

This PR fixes compatibility with Zhipu GLM OpenAI-Compatible APIs.

Currently AstrBot sends:

```json
content: [
  {
    "type": "text",
    "text": "..."
  }
]
```

However GLM expects:

```json
content: "..."
```

which causes the API to return:

```text
messages.content.type 参数非法，取值范围 ['text']
```

## Changes

- Move GLM compatibility logic into `_finally_convert_payload()`.
- Apply conversion only when the model name contains `glm`.
- Convert text-only content arrays into plain strings.
- Restrict the string conversion in `assemble_context()` to GLM only.

## Tested

Tested with:

- `GLM-4.7-Flash`
- OpenAI-Compatible API

The API now works correctly without affecting Gemini, DeepSeek, or other providers.

This patch was generated with assistance from ChatGPT Codex and verified through actual testing.

## Summary by Sourcery

Improve compatibility with Zhipu GLM models in the OpenAI-compatible provider by normalizing chat message content formatting.

Enhancements:
- Normalize GLM chat messages by converting text-only content arrays into plain string content in the final payload conversion.
- Adjust user context assembly to emit plain string content specifically for GLM models when all parts are text-only.

### DIFF
--- a/astrbot/core/provider/sources/openai_source.py
+++ b/astrbot/core/provider/sources/openai_source.py
@@ -992,6 +992,20 @@ class ProviderOpenAIOfficial(Provider):
     def _finally_convert_payload(self, payloads: dict) -> None:
         """Finally convert the payload. Such as think part conversion, tool inject."""
         model = payloads.get("model", "").lower()
+        # GLM's OpenAI-compatible API only accepts string content.
+        if "glm" in model:
+            for message in payloads.get("messages", []):
+                content = message.get("content")
+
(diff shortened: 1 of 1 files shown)


Soulter:
image

seems working well.

The error you encountered may be related to model's modality.

PR #9285, opened 2026-07-14
fix(qqofficial): render @ mentions in group messages

IQ-Director:
- serialize At components as <@openid> markup
- send mention-bearing replies and proactive messages as Markdown
- preserve payload compatibility for media and guild channel messages
- support legacy and current incoming mention formats
- add regression tests for QQ Official @ mentions

QQ Official outgoing messages currently discard `At` components in `_parse_to_qqofficial`, causing user mentions to disappear from replies.

In addition, messages sent through `send_by_session`, including proactive and scheduled group messages, are sent through the plain `content` field. QQ does not correctly resolve `<@openid>` mentions through this path, so the mention may be displayed as plain text instead of an actual `@` mention.

This change preserves `At` components and ensures that messages containing user mentions are sent as Markdown, allowing QQ clients to render them correctly.

### Modifications / 改动点

- Updated `qqofficial_message_event.py` to serialize `At` components as `<@openid>`.
- Skipped `At(qq="all")`, because QQ group chats do not support bot `@all` mentions.
- Ensured event replies containing `At` use Markdown even when Markdown was explicitly disabled on the message chain.
- Updated `send_by_session` so proactive, scheduled, and directly sent group messages containing `At` use `msg_type=2` with a Markdown payload.
- Kept ordinary messages without mentions on the existing plain-text sending path to avoid changing unrelated behavior.
- Preserved compatibility with media messages by removing the Markdown payload when switching to `msg_type=7`.
- Removed the incompatible `msg_type` field when sending messages to guild text channels.
- Added support for cleaning legacy and current incoming mention formats:
  - `<@bot_id>`
  - `<@!bot_id>`
  - `<qqbot-at-user id="bot_id" />`
- Added regression tests for outgoing mention serialization, incoming mention cleanup, and proactive group messages containing `At`.

### Screenshots or Test Results / 运行截图或测试结果

Test command:
uv run --isolated --no-project --with pytest --with pytest-asyncio --with-editable . pytest tests/test_qqofficial_group_message_create.py -q

Test result:
16 passed, 4 warnings in 17.91s

Ruff command:
uv run --isolated --no-project --with ruff ruff check astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py astrbot/core/platform/sources/qqofficial/qqofficial_platform_adapter.py

Ruff result:
All checks passed!

image

## Summary by Sourcery

[...]

PR #9310, opened 2026-07-18
Revert "fix(qqofficial): render @ mentions in group messages"

Soulter:
Reverts AstrBotDevs/AstrBot#9285

## Summary by Sourcery

Revert QQ Official group message @-mention handling and restore previous payload formatting for text and media messages.

Enhancements:
- Simplify QQ Official platform payload construction by always using plain content for base messages and inlining media payload setup instead of helper methods.
- Relax bot mention markup normalization by only stripping specific inline mention patterns instead of generic markup handling.
- Remove special handling of At components when serializing message chains to QQ Official messages, returning to plain-text-only serialization.

Tests:
- Remove tests that validated At component serialization, mention-driven markdown usage, and multiple bot mention markup variants, aligning tests with the reverted behavior.

### DIFF
--- a/astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py
+++ b/astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py
@@ -25,7 +25,7 @@ from tenacity import (
 
 from astrbot.api import logger
 from astrbot.api.event import AstrMessageEvent, MessageChain
-from astrbot.api.message_components import At, File, Image, Plain, Record, Video
+from astrbot.api.message_components import File, Image, Plain, Record, Video
 from astrbot.api.platform import AstrBotMessage, PlatformMetadata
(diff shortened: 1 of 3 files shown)


IQ-Director:
@Soulter Thank you for reviewing and maintaining this feature.

However, according to the QQ Bot official documentation, QQ Official bots do support mentioning users in group chats.

The text-chain documentation explicitly states:

> **@ someone | Available in group chats and text subchannels**

Official documentation:
https://bot.q.qq.com/wiki/develop/api-v2/server-inter/message/trans/text-chain.html

I have also tested the implementation from #9285 locally. The mention was rendered as an actual user mention in a QQ group, rather than plain text, and it worked as expected.

Therefore, it seems inaccurate to say that the QQ Official platform does not support real mentions. 

Could you please reconsider this revert, or her compatibility issue with theimplementation in #9285?

PR #9424, opened 2026-07-28
feat: wire up video input modality for MiniMax providers

octo-patch:
Reason: MiniMax-M3 supports video input, but AstrBot only emitted a textual placeholder for videos and never transmitted the video modality to MiniMax providers.

Changes:
- Populate `ProviderRequest.video_urls` from `Video` attachments (mirroring the existing image flow) so resolved video paths reach the provider instead of only a text placeholder.
- Materialize `video_url` context blocks into data URLs in the OpenAI source payload preparation, so MiniMax via the OpenAI-compatible endpoint transmits video.
- Convert `video_url` blocks into Anthropic `video` blocks in the Anthropic source `_prepare_payload`, so the MiniMax Token Plan provider transmits video.
- Add a `VideoURLPart` content block type for extra user content parts, consistent with `ImageURLPart`/`AudioURLPart`.
- Expose `video` as a model modality option and honor it during request context assembly and context sanitization by modalities.

Checks:
- `ruff check` on all changed files.
- `pytest tests/test_video_input_modality.py` (5 passed).
- `pytest tests/test_anthropic_kimi_code_provider.py tests/test_openai_source.py` (90 passed).

## Summary by Sourcery

Wire video input through the provider pipeline so MiniMax and compatible providers receive actual video modality instead of text placeholders.

New Features:
- Add video URL support to ProviderRequest, including context assembly and logging of video counts.
- Introduce a VideoURLPart content type for user content parts, analogous to image and audio parts, and handle it in OpenAI-style context assembly.
- Expose video as a model modality option and respect it when assembling and sanitizing request contexts across providers.

Enhancements:
- Resolve video attachments and video_url content blocks into data-URL payloads in the OpenAI source so downstream MiniMax-compatible endpoints receive base64 video data.
- Convert OpenAI-style video_url blocks into Anthropic video blocks in the Anthropic source so MiniMax Token Plan provider can consume video input.
- Extend modality-based context sanitization and tool-loop request adjustment to correctly handle unsupported video modality alongside image and audio.

Tests:
- Add tests covering ProviderRequest video_url assembly, modality-based sanitization of video content, OpenAI source video_url materialization, and Anthropic video block conversion.


JosephTian876:
Thanks — mirroring the existing `image_urls` flow rather than inventing a new mechanism is exactly the right shape, and it would be good to see video input land.

I audited this against current `master` while building a plugin-side workaround, and found two gaps that I believe need addressing before it can merge safely. Sharing in case they're useful.

### 1. `gemini_source.py` is untouched → uncaught `KeyError` on every Gemini request

`_prepare_conversation` dispatches user content through a catch-all `else`
(`astrbot/core/provider/sources/gemini_source.py:349-355`):

```python
for item in content:
    if item["type"] == "text":
        parts.append(types.Part.from_text(text=item["text"] or " "))
    elif item["type"] == "image_url":
        parts.append(await process_image_url(item["image_url"]))
    else:
        parts.append(process_audio_url(item["audio_url"]))
```

Once `assemble_context()` emits `{"type": "video_url", ...}` blocks, they take that
`else` branch and raise `KeyError: 'audio_url'` whenever Gemini builds a payload
containing them.

That exception is not caught: `_prepare_conversation` is awaited at `:656`, which is
*outside* the `try:` that starts at `:661`, so the `except APIError` at `:700` never
sees it and it escapes `text_chat`.

Because the block is also persisted into the session history, the failing message is
replayed on every later request in that session — so a single video poisons the
conversation, including plain-text turns that follow. This is the same failure shape
as #9295 (a GIF persisted into history keeps failing long after the triggering
message).

Making the dispatch explicit (`elif item["type"] == "audio_url":` plus a text
fallback for unknown types) would close this.

### 2. Video blocks are never marked transient

The `content_blocks.append({"type": "video_url", ...})` at the end of
`assemble_context()` is a plain dict with no `_no_save` flag.

That user message is appended to `run_context.messages`
(`astrbot/core/agent/runners/tool_loop_agent_runner.py:309-323`) and written back to
the session, so a 20 MB video becomes roughly 27 MB of base64 that is re-sent on
every subsequent request.

`ContentPart.mark_as_temp()` (`astrbot/core/agent/message.py:68`) sets `_no_save`,
and `dump_messages_with_checkpoints` (`:345`) filters those parts out at `:354`.
Applying it to the video part keeps history small; the trade-off is that the model
no longer sees the video on follow-up turns.

### Also

The PR currently reports `mergeable_state: dirty` against `master`, so it needs a
rebase before it can be evaluated.

---

For context: I have a working plugin implementation
([astrbot_plugin_video_vision](https://github.com/JosephTian876/astrbot_plugin_video_vision))
that injects video through `on_llm_request`, tested end-to-end against a real
QQ deployment. Happy to share the measurements (token cost per second of video,
upload-bandwidth limits) or help with a rebase if that would be useful.

PR #9458, opened 2026-07-30
fix(session_waiter): track handler task to prevent execution after timeout

Qixuan112:
## 问题描述

在  第 167 行有一个 TODO 注释，指出需要使用  来跟踪任务，防止超时后 handler 仍然在执行。

## 解决方案

1. **添加任务跟踪**: 在  类中添加  属性来跟踪正在运行的 handler 任务
2. **使用 create_task**: 使用  包装 handler 执行，以便进行任务管理
3. **清理时取消任务**: 在  方法中取消任何正在运行的 handler 任务
4. **防止任务堆积**: 如果新的触发器到达时上一个任务还未完成，则取消上一个任务
5. **正确处理取消**: 显式处理  以避免错误传播

## 改进点

- 防止会话超时后 handler 仍继续执行，造成资源泄漏
- 确保在会话结束时正确清理所有相关任务
- 避免并发触发导致的多个 handler 同时执行

## 测试

- ✅ Python 语法检查通过
- ✅ 代码与现有使用场景兼容
- ✅ 保持了原有的异常处理逻辑

## 相关 TODO

解决了  的 TODO 注释。

## Summary by Sourcery

Track and manage session handler tasks to prevent handlers from running after session timeout.

Bug Fixes:
- Ensure session handlers do not continue executing after a session timeout, avoiding resource leaks and unintended processing.

Enhancements:
- Add per-session handler task tracking and cancellation to avoid overlapping handler executions and improve cleanup behavior on session end.

### DIFF
--- a/astrbot/core/utils/session_waiter.py
+++ b/astrbot/core/utils/session_waiter.py
@@ -121,6 +121,9 @@ class SessionWaiter:
         self._lock = asyncio.Lock()
         """需要保证一个 session 同时只有一个 trigger"""
 
+        self._handler_task: asyncio.Task | None = None
+        """当前正在运行的 handler 任务"""
+
     async def register_wait(
         self,
         handler: Callable[[SessionController, AstrMessageEvent], Awaitable[Any]],
@@ -148,6 +151,9 @@ class SessionWaiter:
             FILTERS.remove(self.session_filter)
         except ValueError:
             pass
+        # 取消正在运行的 handler 任务
+        if self._handler_task and not self._handler_task.done():
+            self._handler_task.cancel()
         self.session_controller.stop(error)
 
     @classmethod
@@ -164,9 +170,24 @@ class SessionWaiter:
                         [copy.deepcopy(comp) for comp in event.get_messages()],
                     )
                 try:
-                    # TODO: 这里使用 create_task，跟踪 task，防止超时后这里 handler 仍然在执行
+                    # 使用 create_task 跟踪任务，防止超时后 handler 仍然在执行
                     assert session.handler is not None
-                    await session.handler(session.session_controller, event)
+
+                    async def _run_handler():
+                        try:
+                            await session.handler(session.session_controller, event)
+                        except Exception as e:
+                            session.session_controller.stop(e)
+
+                    # 取消上一个可能还在运行的任务
+                    if session._handler_task and not session._handler_task.done():
+                        session._handler_task.cancel()
+
+                    session._handler_task = asyncio.create_task(_run_handler())
+                    await session._handler_task
+                except asyncio.CancelledError:
+                    # 任务被取消时不需要处理
+                    pass
                 except Exception as e:
                     session.session_controller.stop(e)
 
--- a/astrbot/core/utils/session_waiter.py
+++ b/astrbot/core/utils/session_waiter.py
@@ -139,21 +139,22 @@ class SessionWaiter:
         try:
             return await self.session_controller.future
         except Exception as e:
-            self._cleanup(e)
+            await self._cleanup(e)
             raise e
         finally:
-            self._cleanup()
+            await self._cleanup()
 
-    def _cleanup(self, error: Exception | None = None) -> None:
+    async def _cleanup(self, error: Exception | None = None) -> None:
         """清理会话"""
         USER_SESSIONS.pop(self.session_id, None)
         try:
             FILTERS.remove(self.session_filter)
         except ValueError:
             pass
-        # 取消正在运行的 handler 任务
-        if self._handler_task and not self._handler_task.done():
-            self._handler_task.cancel()
+        # 使用锁保护任务取消操作，防止与 trigger 方法竞态
+        async with self._lock:
+            if self._handler_task and not self._handler_task.done():
+                self._handler_task.cancel()
         self.session_controller.stop(error)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


Qixuan112:
Fixed by re-raising CancelledError in _run_handler so cancellations are handled by the outer except block instead of being treated as errors.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

xiaoyuyu6420 (commented):
## Review

This PR addresses the TODO at line 217 — the original code `await session.handler(...)` inside `_lock` means a long-running handler blocks the lock, and after timeout cleanup the handler keeps running with no way to cancel it. Tracking the task is the right direction.

### Strengths

1. **Waiting outside the lock** (`task_to_await` pattern) — correctly avoids holding `_lock` during handler execution, which prevents deadlock when `_cleanup()` tries to acquire the same lock to cancel the task.
2. **Re-raising `CancelledError`** inside `_run_handler` — correct, ensures the outer except clause can distinguish cancellation from other exceptions.
3. **Cancelling previous task before creating new one** — prevents overlapping handler execution if `trigger()` is called rapidly.

### Issues

**1. `_cleanup()` is now `async` — this is a breaking change for any caller that invokes it synchronously**

`_cleanup()` was a sync method. Making it `async` means any code that calls `self._cleanup()` without `await` will silently create a coroutine that never executes. Within this file both call sites are updated, but this is a public-ish method on `SessionWaiter` — third-party code or plugins that call `_cleanup()` directly will break silently.

**2. Deadlock risk: `_cleanup()` acquires `_lock`, but `trigger()` also holds `_lock` when the handler runs**

Consider this sequence:
1. `trigger()` acquires `_lock`, creates `_handler_task`, releases `_lock`, then `await task_to_await`
2. While handler is running, timeout fires → `register_wait()` calls `await self._cleanup()`
3. `_cleanup()` tries `async with self._lock` to cancel `_handler_task`
4. If the handler itself calls `controller.keep()` (which doesn't touch `_lock`), this works. But if the handler somehow triggers another `trigger()` call on the same session (recursive), `_lock` would already be released by then, so this is probably OK.

Actually, on closer inspection, since `trigger()` releases `_lock` before awaiting `task_to_await`, and `_cleanup()` acquires `_lock` only to cancel the task, there shouldn't be a deadlock. But the interaction is subtle enough to warrant a comment explaining why it's safe.

**3. No test for the task cancellation behavior**

The core value of this PR — cancelling a handler after timeout — has no test. A test should verify:
- When timeout fires while handler is running, `_handler_task` is cancelled
- When `trigger()` is called again before previous handler finishes, previous task is cancelled

**4. `_handler_task` exception handling asymmetry**

In `_run_handler`, exceptions are caught and passed to `session.session_controller.stop(e)`. But in the outer `await task_to_await`, only `CancelledError` is caught — if the task raises any other exception (which shouldn't happen because `_run_handler` catches them all), it would propagate to `trigger()`'s caller. This is probably fine but the double-layer exception handling is confusing.

### Summary

Good direction — the task tracking pattern correctly addresses the TODO. The main concerns are: (1) the `async _cleanup()` breaking change, (2) missing tests for the cancellation behavior, and (3) the subtle lock interaction deserves documentation. Consider adding a test that verifies a timed-out handler actually gets cancelled.

Qixuan112:
@xiaoyuyu6420 感谢详细审查。已逐项处理并推送：

### 1. `_cleanup()` async → 保留，已文档化
`_cleanup` 以 `_` 前缀命名，全仓库 grep 确认无外部调用者。保持 async 是因为需要在 `_lock` 保护下取消 `_handler_task`（避免与 `trigger` 竞态）。已在 docstring 说明其为内部私有 async 方法。

### 2. 锁交互安全 → 已补注释
在 `trigger`（锁外 await）和 `_cleanup`（锁内 cancel）两处补充了不会死锁的说明：`trigger` 锁内只做同步操作、不在锁内 await handler；`_cleanup` 仅在锁内取消任务；两者不会互锁。

### 3. 单元测试 → 已补齐
新增 `tests/unit/test_session_waiter.py`，6 个用例覆盖：
- 基础流程（trigger → handler → register_wait）
- 重复 trigger 取消上一个 handler
- `_cleanup` 取消运行中 handler
- `CancelledError` 不流入 `stop()` — cleanup + double-trigger 两路径
- handler 普通异常经 `stop(e)` 传播
全部 6 用例稳定通过（strict asyncio 模式，3 次重跑无 flaky）。

### 4. 双层异常处理 → 已补注释
分别在 `_run_handler` 内层（普通异常 → `stop(e)`；`CancelledError` re-raise）和外层 `await task_to_await`（静默处理取消）加了注释说明分工。

### 5. `_handler_task` 引用清理
在 `task_to_await` 完成后用 `finally` 块安全清理引用（`if session._handler_task is task_to_await` 守卫），防止误清并发 trigger 的新任务。

另外注意到 2 个 PR 之前就存在的问题（非本次范围）：`register_wait` 中 `except` 和 `finally` 都调 `_cleanup`（当前幂等故无害），以及 `_run_handler` 中 `stop(e)` 自身若抛异常会逃逸到外层（当前不可达）。这两项可留待后续。

_🤖 Prepared by Claude Code_

xiaoyuyu6420:
复审确认：保留 `_cleanup` 为 async 的理由成立——锁内取消 `_handler_task` 才能避免与 `trigger` 的竞态，docstring 已把私有契约说清楚；两处锁交互的死锁排除说明也补上了；`tests/unit/test_session_waiter.py` 6 个用例覆盖了基础流程、重复 trigger 取消和 `_cleanup` 取消运行中 handler 的场景。后续把 `CancelledError` 在 `_run_handler` 里 re-raise 而非按错误处理，也是正确的方向。没有进一步意见，同意合入。

PR #9518, opened 2026-08-02
fix: warn when an explicitly configured WebUI dist is stale

JosephTian876:
## Problem

`resolve_dashboard_dist()` returns an explicitly configured `--webui-dir` immediately:

```python
if webui_dir and Path(webui_dir).exists():
    return Path(webui_dir).absolute()
```

Every other branch of the same function verifies the dist against the running core, and two of them already warn when it does not match:

| Branch | Version checked? |
| --- | --- |
| `data/dist` compatible | yes, `_is_dist_compatible` |
| bundled dist | yes, plus an info log |
| `data/dist` version mismatch | yes, warns and names both versions |
| `data/dist` incomplete | yes, warns |
| **explicit `--webui-dir`** | **no check at all** |

So the explicit directory is the only way to end up serving a dashboard that does not match the running core with no diagnostic whatsoever.

## How I hit it

I run the packaged desktop distribution, which launches the backend with `--webui-dir`. Updating took the core from 4.26.8 to 4.27.0, but the launcher kept its bundled WebUI directory untouched, so the dashboard stayed on the 4.26.8 build. The UI simply did not have the new extension/persona management pages, and there was nothing in the log to explain it — the update had reported success.

It took a byte-level comparison against `AstrBot-v4.27.0-dashboard.zip` to establish what had happened:

```
served dist   : index-8HR-5XWP.js, 569 assets, no assets/version
v4.27.0 build : index-Dez8SO42.js, 243 assets, assets/version = v4.27.0
```

Five identifiers that exist in the v4.27.0 dashboard build and not in v4.26.8 (`capability-source-chip`, `CapabilitySourceChip`, `PluginWorkspacePage`, `McpServersPage`, `PersonaCapabilitiesEditor`) were absent from the served assets.

The launcher is of course the component that should have replaced those files. But the core already knows how to detect this — `_read_dashboard_version()` and `_is_dist_compatible()` are right there and used by every other branch — so silently serving mismatched assets is a gap worth closing regardless of who packaged them.

## Change

Check the explicit directory the same way as the others and warn on a mismatch, reusing the wording of the existing stale-`data/dist` warning.

**Behaviour is unchanged** — the directory is still served, since refusing to serve a stale dashboard would be worse than serving one. It is only no longer silent:

```
Serving the explicitly configured WebUI directory even though it does not
declare a version matching core: unknown, expected v4.27.0 (D:\AstrBot\webui).
Some dashboard features may not work until matching assets are available.
```

`resolve_dashboard_dist()` is called once, from `DashboardServer.__init__`, so this is a single line at startup.

## A note on the "unknown" case

`assets/version` is written by the release workflows, not by `pnpm run build`, so a locally built dashboard passed via `--webui-dir` has no marker and will produce this warning with `unknown`. I think that is acceptable — it is one startup line, it names the directory, and it is accurate — but if you would rather keep local development silent I am happy to downgrade the missing-marker case to `info` and keep `warning` only for a marker that is present and different.

## Tests

Adds `tests/unit/test_dashboard_dist_resolution.py` — 6 cases: matching version stays quiet, mismatched version warns, missing marker warns as `unknown`, and three cases that must not warn (non-existent path, `None`, empty string).

- Reverting the `dashboard_assets.py` change makes exactly the 2 warning cases fail; the 4 must-not-warn guards pass either way.
- `tests/unit`: 799 passed, 6 failed — the same 6 fail on unmodified `master` here (Windows-local shell/python execution, video attachment and sandbox path tests), unrelated to this change.
- `ruff format --check .` and `ruff check .` both pass.

## Related

#9484 addresses the same shape of problem in `ProviderManager.get_using_provider()`: an existing warning that cannot fire in the case that actually matters, so a substitution happens silently.

## Summary by Sourcery

Warn on mismatched WebUI assets when an explicit dashboard directory is configured and add coverage for this behavior.

Enhancements:
- Validate explicitly configured WebUI directories against the core version and log a warning when the dashboard assets appear stale or unverifiable.

Tests:
- Add unit tests covering explicit WebUI directory resolution, including matching, mismatched, missing-version, and non-existent directory scenarios.

### DIFF
--- a/astrbot/core/dashboard_assets.py
+++ b/astrbot/core/dashboard_assets.py
@@ -141,15 +141,16 @@ def resolve_dashboard_dist(webui_dir: str | Path | None = None) -> Path | None:
         Explicit, managed, bundled, or stale fallback dist in priority order;
         None when an existing managed dist is incomplete.
     """
-    if webui_dir and Path(webui_dir).exists():
-        explicit_dist = Path(webui_dir).absolute()
+    explicit_dist = Path(webui_dir).absolute() if webui_dir else None
+    if explicit_dist is not None and explicit_dist.exists():
         if not _is_dist_compatible(explicit_dist, VERSION):
+            explicit_version = _read_dashboard_version(explicit_dist) or "unknown"
             logger.warning(
                 "Serving the explicitly configured WebUI directory even though it "
                 "does not declare a version matching core: %s, expected v%s (%s). "
                 "Some dashboard features may not work until matching assets are "
                 "available.",
-                _read_dashboard_version(explicit_dist) or "unknown",
+                explicit_version,
                 VERSION,
                 explicit_dist,
             )
--- a/astrbot/core/dashboard_assets.py
+++ b/astrbot/core/dashboard_assets.py
@@ -142,7 +142,18 @@ def resolve_dashboard_dist(webui_dir: str | Path | None = None) -> Path | None:
         None when an existing managed dist is incomplete.
     """
     if webui_dir and Path(webui_dir).exists():
-        return Path(webui_dir).absolute()
+        explicit_dist = Path(webui_dir).absolute()
+        if not _is_dist_compatible(explicit_dist, VERSION):
+            logger.warning(
+                "Serving the explicitly configured WebUI directory even though it "
+                "does not declare a version matching core: %s, expected v%s (%s). "
+                "Some dashboard features may not work until matching assets are "
+                "available.",
+                _read_dashboard_version(explicit_dist) or "unknown",
+                VERSION,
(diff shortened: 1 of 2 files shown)


JosephTian876:
Both taken in 1959b86f — thanks.

`Path(webui_dir)` is now built once (that double construction predated this PR, so good catch on it), and the version read for the warning is held in a local.

Re-verified after the refactor: reverting `dashboard_assets.py` to `master` still fails exactly the two warning cases and passes the four must-not-warn guards, so the tests are still anchored to the behaviour and not to the shape of the code.

Soulter (approved):

PR #9615, opened 2026-08-10
fix(qqofficial): send proactively before passive reply window expires

TheRainstorm:
## Problem

QQ official passive replies (`msg_id`-based) are only valid for **5 minutes**. Long-running tasks (e.g. an agent running tools for several minutes) exceed this window, and the final reply fails with:

```
botpy.errors.ServerError: 回复消息msg_id已过期
```

The existing fallback then retries once without `msg_id`, wasting a request and often failing again for group messages.

## Fix

When a group reply's age approaches the 5-minute passive window (60s margin), build the payload **without** `msg_id` from the start, i.e. send via the proactive (active-message) API directly:

- `_QQOFFICIAL_PASSIVE_REPLY_SECONDS = 300`, `_QQOFFICIAL_PROACTIVE_MARGIN_SECONDS = 60`
- `force_proactive` computed in `_post_send_one` for `GroupMessage`

C2C/other sources keep the existing `msg_id` behavior; the `msg_id`-less proactive retry in `_send_with_markdown_fallback` remains as a safety net.

## Notes

- Active group messages use the platform's proactive message quota; this only kicks in for long tasks near the 5-minute boundary, so normal replies are unaffected.

## Summary by Sourcery

Bug Fixes:
- Prevent failures of long-running QQ official group replies by avoiding expired msg_id-based passive replies near the 5-minute limit.

### DIFF
--- a/astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py
+++ b/astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py
@@ -4,6 +4,7 @@ import copy
 import logging
 import os
 import random
+import time
 from typing import cast
 
 import aiofiles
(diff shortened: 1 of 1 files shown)


TheRainstorm:
Addressed the review feedback:

- **Monotonic clock**: added `created_at_monotonic = time.monotonic()` at event construction and switched the passive-window check to `time.monotonic()`, so wall-clock changes during long-running tasks no longer affect the decision.
- **Configurable constants**: kept `_QQOFFICIAL_PASSIVE_REPLY_SECONDS`/`_QQOFFICIAL_PROACTIVE_MARGIN_SECONDS` as module constants to match the file's existing style and avoid new config plumbing for a niche value; happy to make them configurable if maintainers prefer.

PR #9616, opened 2026-08-10
feat(qqofficial): chunked upload for large media and text fallback

TheRainstorm:
## Problem

QQ v2 inline uploads (`file_data` base64 in `POST /v2/groups/{group_openid}/files`) are capped around **10MB**. Larger local files (e.g. an 18MB xlsx produced by a long task) fail with:

```
[botpy] HTTP 413 Request Entity Too Large
```

and the bot ends up sending nothing (the subsequent `msg_type=2` payload with empty markdown content fails with `40034011 无效 markdown content`).

### 1. Chunked upload for large local files
When `file_source` is a local file \> 10MB, use the official chunked flow instead of base64:

1. `POST .../upload_prepare` with `md5`/`sha1`/`md5_10m` → `upload_id` + `block_size` + presigned COS part URLs
2. `PUT` each part to its presigned URL (concurrency follows server `upload_config`, capped at 4; each part retried)
3. `POST .../upload_part_finish` per part (retries on `biz_code 40093001`)
4. `POST .../files` with `upload_id` to merge → `file_info`

Daily cumulative quota (`biz_code 40093002`) surfaces a friendly message instead of a raw error. C2C (`/v2/users/{openid}`) and group (`/v2/groups/{group_openid}`) endpoints are both supported.

### 2. Plain-text fallback so users never get silence
- If a markdown or media send fails (e.g. `40034011` invalid markdown content), retry as `msg_type=0` `content=`.
- If a media-only message fails with no text attached, send a short text explanation instead of nothing.
- Media upload failures raise `QQMediaUploadError` and degrade the payload to plain text.

## Testing

- Mocked end-to-end test of the chunked driver on a 17MB file: parts reconstruct the source byte-for-byte, md5/block_size match, `upload_part_finish` 40093001 retry works, daily-limit error surfaces correctly.
- Fallback unit tests for 40034011 → content-only, media-only failure → explanation text.
- `ruff check` + `ruff format` clean.

## Summary by Sourcery

Implement chunked upload for large QQ media files and introduce robust text fallbacks when media or markdown message delivery fails.

New Features:
- Support QQ official chunked upload flow for large local media files exceeding the inline upload size limit.
- Provide automatic plain-text fallback when markdown or media messages fail to send so users still receive a response.

Enhancements:
- Introduce dedicated QQ media and API error types and stricter response validation to surface clearer upload failures and quotas.
- Add centralized QQ API helper and chunk-part upload orchestration with concurrency control and retries for more reliable media delivery.

### DIFF
--- a/astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py
+++ b/astrbot/core/platform/sources/qqofficial/qqofficial_message_event.py
@@ -1,12 +1,14 @@
 import asyncio
 import base64
 import copy
+import hashlib
 import logging
 import os
 import random
 from typing import cast
 
 import aiofiles
+import aiohttp
 import botpy
 import botpy.errors
 import botpy.message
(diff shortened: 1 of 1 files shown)


TheRainstorm:
Addressed the review feedback:

1. **part_index validation** (bug_risk): `_upload_one_part` now parses `part_index` defensively and raises a clear `QQMediaUploadError` (with the truncated part payload) when it is missing, non-numeric, or `<= 0`, before any offset math.
2. **Centralized degradation**: `_send_with_markdown_fallback` now reuses `_degrade_media_payload_to_text` for both fallback paths, so stripping `markdown`/`media`, setting `msg_type`/`content`, stream newline handling, and the default explanation text live in one place.
3. **Chunked session object**: introduced `_ChunkUploadSession` to carry shared state and slimmed `_upload_one_part`/`_part_finish_with_retry` signatures.
4. **botpy internals**: added a docstring note explaining that the raw botpy HTTP session is used because botpy only surfaces the API error message and drops the platform error code needed for `40093001`/`40093002` handling; this matches the existing `self.bot.api._http.request` pattern already used in this file.

Re-ran the mocked end-to-end tests (17MB file: parts reconstruct byte-for-byte, part_finish 40093001 retry, daily-limit message, malformed part_index) and the fallback tests — all pass; `ruff check` + `ruff format` clean.

Soulter:
Thanks for pioneering the QQ Official chunked-upload implementation and documenting the protocol details. I opened #9646 with what I believe is a cleaner integration approach: the multipart driver now lives in a dedicated module, C2C and group entry points remain explicit, unrelated message-fallback behavior is kept out of scope, and the implementation is rebased on the media error-propagation behavior already merged into `master`. It also handles both zero-based documented part indexes and one-based indexes observed at runtime.

Your work is credited in the commit with `Co-authored-by: Fuyan Yuan <33221728+TheRainstorm@users.noreply.github.com>`.

PR #9769, opened 2026-08-22
fix: support anthropic SDK 1.0 httpx2 module

wcqqq1214:
### Modifications / 改动点

Anthropic SDK 1.0.0 (released 2026-08-20; see [Claude Platform release notes](https://platform.claude.com/docs/en/release-notes/overview#august-20-2026)) renamed its internal transport module from `anthropic._base_client.httpx` to `httpx2`. Since `uv.lock` is gitignored, CI resolves `anthropic>=0.51.0` to the latest PyPI release, so since 2026-08-21 every run installs `anthropic==1.0.0` and the existing unit test `test_create_http_client_uses_anthropic_httpx_module` fails (`module 'anthropic._base_client' has no attribute 'httpx'`).

- `astrbot/core/provider/sources/anthropic_source.py`: resolve the SDK's own httpx module preferring `httpx2` (anthropic 1.x) and falling back to `httpx` (anthropic 0.x), so proxy clients still pass the SDK's `isinstance` check when multiple httpx installations are on `sys.path`.
- `tests/test_anthropic_kimi_code_provider.py`: mirror the same resolution in the test instead of hard-coding the private `httpx` attribute.

- [x] This is NOT a breaking change. / 这不是一个破坏性变更。

### Verification Steps / 验证步骤

1. Ran with `anthropic==1.0.0` installed (venv, `httpx2` present, `httpx` absent):
   - `pytest tests/test_anthropic_kimi_code_provider.py` → 27 passed
   - full `pytest tests/` → 2163 passed
   - `ruff format --check` and `ruff check` → clean
2. Fallback path (`httpx` attribute) is covered by the existing `test_create_http_client_falls_back_to_global_httpx_module`.

### Screenshots or Test Results / 运行截图或测试结果

```text
$ pytest tests/test_anthropic_kimi_code_provider.py -q
27 passed, 1 warning in 1.06s

$ pytest tests/ -q
2163 passed, 19 warnings in 74.88s
```

---

### Checklist / 检查清单

- [x] 😊 If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc.
  / 如果 PR 中有新加入的功能，已经通过 Issue / 邮件等方式和作者讨论过。

- [x] 👀 My changes have been well-tested, **and "Verification Steps" and "Screenshots" have been provided above**.
  / 我的更改经过了良好的测试，**并已在上方提供了“验证步骤”和“运行截图”**。

[...]

PR #9786, opened 2026-08-23
fix: accumulate all text blocks in Anthropic non-streaming response

AmirF194:
`ProviderAnthropic._query` (the non-streaming path in `anthropic_source.py`) sets `llm_response.completion_text` inside a loop over `completion.content`, once per text content block, instead of appending. When a `Message` carries more than one text block, only the last one survives; every earlier text block is silently dropped from the reply and from conversation history. This can happen with the Citations API and with extended-thinking flows, both of which can split a response into multiple text blocks. The streaming sibling, `_query_stream`, already accumulates correctly with `final_text += event.delta.text`.

### Modifications / 改动点

Collect each text block into a list during the loop and join them once after the loop, instead of overwriting `completion_text` per block. `thinking` and `tool_use` handling are unchanged.

Added `test_query_accumulates_multiple_text_blocks`, which feeds `ProviderAnthropic._query` a fake `Message` with two text blocks. It fails on unmodified `master` (`AssertionError`, only the second block's text survives) and passes with this fix. The full `tests/test_anthropic_kimi_code_provider.py` file: 28 passed, 1 pre-existing failure unrelated to this change (`test_create_http_client_uses_anthropic_httpx_module`, an SDK version drift already being fixed in open PR #9769). `ruff format --check` and `ruff check` are clean on both changed files.

Not verified: a live call to the provider's API. This was reproduced and tested against the SDK's `Message`/content-block shape directly, not against a real multi-block response from the API.

---

### Checklist / 检查清单

- [x] If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc.

- [x] My changes have been well-tested, **and "Verification Steps" and "Screenshots" have been provided above**.
  / 我的更改经过了良好的测试，**并已在上方提供了"验证步骤"和"运行截图"**。

- [x] I have ensured that no new dependencies are introduced, OR if new dependencies are introduced, they have been added to the appropriate locations in `requirements.txt` and `pyproject.toml`.

  / 我的更改没有引入恶意代码。

## Summary by Sourcery

Accumulate every text block returned by Anthropic’s non-streaming API so complete responses and conversation history are preserved.

Bug Fixes:
- Preserve all text content blocks in Anthropic non-streaming responses instead of returning only the final block.

Tests:
- Add coverage verifying that multiple Anthropic text blocks are concatenated in the response.

### DIFF
--- a/astrbot/core/provider/sources/anthropic_source.py
+++ b/astrbot/core/provider/sources/anthropic_source.py
@@ -540,11 +540,11 @@ class ProviderAnthropic(Provider):
             )
 
         llm_response = LLMResponse(role="assistant")
+        completion_text_parts: list[str] = []
 
         for content_block in completion.content:
             if content_block.type == "text":
-                completion_text = str(content_block.text).strip()
-                llm_response.completion_text = completion_text
+                completion_text_parts.append(str(content_block.text))
 
             if content_block.type == "thinking":
                 reasoning_content = str(content_block.thinking).strip()
@@ -556,6 +556,7 @@ class ProviderAnthropic(Provider):
                 llm_response.tools_call_name.append(content_block.name)
                 llm_response.tools_call_ids.append(content_block.id)
 
+        llm_response.completion_text = "".join(completion_text_parts).strip()
         llm_response.id = completion.id
         llm_response.usage = self._extract_usage(completion.usage)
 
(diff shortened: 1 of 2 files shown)


AmirF194:
Just checking in on this one, no rush at all. Happy to rebase or split the fix if that makes review easier.

AmirF194:
Closing this for now since it's had no review in a month. I'll reopen if there's interest later.

PR #9793, opened 2026-08-24
fix(aiocqhttp): evict stale reverse websocket connections

jmt059:
## Summary

- add a configurable inbound-frame timeout for aiocqhttp reverse WebSocket connections
- close stale connections so the OneBot client can reconnect after a half-open network failure
- replace older same-account connections without allowing their cleanup path to remove the new API mapping
- add focused tests for timeout, disabled timeout, close compatibility, replacement, and cleanup races

## Motivation

After a transient network or NAT interruption, a reverse WebSocket can remain TCP-established while no longer delivering application frames. aiocqhttp 1.4.4 waits indefinitely in `websocket.receive()`, so the adapter cannot distinguish this state from a healthy idle connection. Restarting the adapter recovers the connection because it closes the stale socket, but there is no automatic recovery path.

The new `ws_reverse_idle_timeout` setting defaults to 60 seconds and can be set to `0` to disable the guard. In the validated NapCat setup, heartbeat and reconnect intervals are both 5 seconds, so the default requires roughly 12 consecutive missed heartbeat frames before eviction.

## Verification

Validated against AstrBot master commit `c6a14e0600485293bd88cf78c04ecec967b21b50`:

```text
python -m ruff check <changed files>                         passed
python -m ruff format --check <changed files>                passed
python -m compileall -q <changed files>                      passed
python -m pytest -q tests/unit/test_aiocqhttp_websocket_guard.py
8 passed
```

No new dependencies are introduced.

## Checklist

- [x] This is not a breaking API change.
- [x] The change has focused automated tests.
- [x] No new dependency is introduced.
- [x] Logs contain the adapter label and no message contents or account identifiers.

## Summary by Sourcery

Prevent stale aiocqhttp reverse WebSocket connections from blocking automatic client recovery.

New Features:
- Add a configurable idle timeout for inbound reverse WebSocket frames, with a default of 60 seconds and an option to disable it.

Bug Fixes:
- Automatically evict stale reverse WebSocket connections after prolonged inactivity so clients can reconnect after half-open network failures.
- Safely replace older same-account connections without allowing their cleanup to remove the active connection mapping.

Enhancements:
- Improve WebSocket closing compatibility across asynchronous and synchronous close implementations.

Tests:
- Add focused unit coverage for idle timeout behavior, disabled timeouts, close compatibility, connection replacement, and cleanup races.

### DIFF
--- a/astrbot/core/config/default.py
+++ b/astrbot/core/config/default.py
@@ -362,6 +362,7 @@ CONFIG_METADATA_2 = {
                         "ws_reverse_host": "0.0.0.0",
                         "ws_reverse_port": 6199,
                         "ws_reverse_token": "",
+                        "ws_reverse_idle_timeout": 60,
                     },
                     "个人微信": {
                         "id": "weixin_personal",
@@ -824,6 +825,11 @@ CONFIG_METADATA_2 = {
                         "description": "反向 Websocket Token",
                         "type": "string",
                         "hint": "反向 Websocket Token。未设置则不启用 Token 验证。",
+                    },
+                    "ws_reverse_idle_timeout": {
+                        "description": "反向 Websocket 入站帧超时",
+                        "type": "int",
+                        "hint": "超过该秒数未收到任何帧时关闭连接并等待协议端重连；设为 0 可禁用。",
                     },
                     "wecom_ai_bot_name": {
                         "description": "企业微信智能机器人的名字",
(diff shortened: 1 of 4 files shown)


jmt059:
Closing for now. This transport-level change has not yet been validated in a production deployment. I will only reopen or submit a revised PR after addressing the known race and compatibility concerns and completing real-world validation.

PR #9915, opened 2026-09-02
fix: unregister plugin web APIs on plugin unload

lxfight:
### Motivation / 动机

`Context.registered_web_apis` never released entries registered by plugins. After a plugin was uninstalled, disabled or reloaded, its registered web API handlers stayed in the list forever:

- **Memory leak**: the list holds bound methods of plugin instances, so purged modules cannot be garbage collected. Repeated install/uninstall cycles keep growing the list.
- **Ghost routes**: requests to `/api/v1/plugins/extensions/<uninstalled-plugin>/...` still match the stale handler and execute it, instead of returning 404.
- **Reload residue**: `register_web_api` only replaces entries with the exact same route and methods; routes removed or changed in a new plugin version stay forever.

`PluginManager._unbind_plugin()` already cleans up event handlers, LLM tools, platform adapters and `sys.modules` — this PR closes the gap by also releasing registered web APIs.

### Modifications / 改动点

- `astrbot/core/star/context.py`: add `Context.unregister_web_apis(module_path)` classmethod that removes all web APIs whose handler is owned by the given plugin module path prefix (checked via the handler function module and, for bound methods, the owner class module).
- `astrbot/core/star/star_manager.py`:
  - `_unbind_plugin()`: unregister web APIs owned by the plugin (shared cleanup path for uninstall, disable and single-plugin reload).
  - `_cleanup_plugin_state()`: unregister web APIs for plugins that failed to load.
  - `reload(all)`: clear all registered web APIs alongside the existing registry clears, since every plugin is re-loaded afterwards.
- `tests/test_plugin_manager.py`: unit tests for ownership-prefixed removal (other plugins and core routes are preserved) and an integration test for `_unbind_plugin()`.

```
$ uv run pytest tests/test_plugin_manager.py tests/test_dashboard.py tests/test_fastapi_v1_dashboard.py -q
64 passed (plugin manager, includes 2 new tests)
172 passed (dashboard regression)
```

Verification steps:

1. Install a plugin that calls `context.register_web_api(...)`.
2. Uninstall it, then request its extension endpoint — it now returns 未找到该路由 instead of executing the stale handler.
3. Check logs for `Removed N registered web API(s) from plugin ...`.

---

### Checklist / 检查清单

- [ ] 😊 If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc. 

  / 我的更改没有引入恶意代码。

## Summary by Sourcery

Ensure plugin web APIs are fully removed whenever plugin state is cleaned up so unloaded plugins no longer retain memory or serve ghost routes.

Bug Fixes:
- Remove plugin-owned web API registrations when plugins are unloaded, disabled, reloaded, or fail to load, preventing stale routes, memory leaks, and handlers from removed plugin versions from persisting.

Enhancements:
- Preserve web APIs belonging to other plugins and core components while identifying ownership across plugin submodules and bound methods.

Tests:
- Add unit and integration coverage for selective web API cleanup during plugin unbinding.

### DIFF
--- a/astrbot/core/star/context.py
+++ b/astrbot/core/star/context.py
@@ -726,6 +726,51 @@ class Context:
                 return
         self.registered_web_apis.append((route, view_handler, methods, desc))
 
+    @classmethod
+    def unregister_web_apis(cls, module_path: str) -> int:
+        """Unregister all web APIs registered by a plugin.
+
+        Args:
+            module_path: Plugin module path prefix, e.g. ``data.plugins.my_plugin``.
+            Handlers whose defining module equals this path or resides under it
+            are removed, so that unloading a plugin releases its handlers and
+            avoids ghost routes.
+
+        Returns:
+            The number of web API registrations removed.
+        """
+        if not module_path:
+            return 0
+        prefix = f"{module_path}."
+        remaining: list[RegisteredWebApi] = []
+        removed = 0
+        for api in cls.registered_web_apis:
+            handler = api[1]
+            handler_modules = set()
+            handler_module = getattr(handler, "__module__", None)
+            if isinstance(handler_module, str):
+                handler_modules.add(handler_module)
+            # Bound methods also expose the owner class module, which may
+            # differ from the function module in edge cases.
+            owner = getattr(handler, "__self__", None)
+            if owner is not None:
+                class_module = type(owner).__module__
+                if isinstance(class_module, str):
+                    handler_modules.add(class_module)
+            if any(
+                module == module_path or module.startswith(prefix)
+                for module in handler_modules
+            ):
+                removed += 1
+                logger.debug(
+                    f"Removed registered web API {api[0]} owned by {module_path}"
+                )
+                continue
+            remaining.append(api)
+        if removed:
+            cls.registered_web_apis[:] = remaining
+        return removed
+
     """
     以下的方法已经不推荐使用。请从 AstrBot 文档查看更好的注册方式。
     """
--- a/astrbot/core/star/star_manager.py
+++ b/astrbot/core/star/star_manager.py
@@ -807,6 +807,13 @@ class PluginManager:
                 llm_tools.func_list.remove(tool)
                 logger.info(f"Removed tool: {tool.name}")
 
+        # 清理插件注册的 Web API
+        removed_web_apis = Context.unregister_web_apis(module_prefix)
+        if removed_web_apis:
+            logger.info(
+                f"Removed {removed_web_apis} registered web API(s) from plugin {dir_name}",
+            )
+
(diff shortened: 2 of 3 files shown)


lxfight:
Regarding the Sourcery finding about `turn_off_plugin()` not calling `_unbind_plugin()`: this is an intentional split across two PRs, not an oversight.

- **Disable keeps entries registered by design.** Disabling is a lightweight toggle: the plugin module, handlers and instance stay loaded so re-enabling is cheap. Symmetrically, static Plugin Pages also keep files on disk when a plugin is disabled and gate them at request time via the `activated` check (403). Unregistering web APIs on disable would change the response semantics from "plugin disabled" to "route not found", which is less accurate, and would make the activation gate dead code.
- **The request-time gate lands in #9916**: `_call_plugin_extension()` resolves the owning plugin from the handler module and rejects calls when it is disabled, aligning the dynamic API with the static pages behavior.
- Memory-wise there is no leak in the disabled state: the plugin instance is intentionally kept alive while disabled (same as its handlers); #9915 fixes the case where the plugin is *gone* (uninstalled/reloaded) while its bound handlers were retained forever.

Soulter:
Reviewed head `61785b6635bc8c8a2d0b1ab1b1e38de4700a19ec`. One cleanup gap remains:

**[P2] Unwrap `functools.partial` before resolving handler ownership** (`astrbot/core/star/context.py:750–759`). `register_web_api()` accepts an async callable, including `functools.partial(plugin_handler, ...)`. For a partial, `__module__` is `functools` and `__self__` is absent, so `unregister_web_apis()` never reaches the underlying plugin function or bound method. I reproduced this by registering a partial of an async function defined in `data.plugins.demo_plugin.main`, then calling `_unbind_plugin()`: the route remained in `Context.registered_web_apis`, `_match_registered_web_api()` still returned it, and awaiting the handler still executed the unloaded plugin's code. This leaves the ghost-route/reference-retention issue unresolved for these handlers.

Please unwrap partials before checking the underlying callable's module/owner, and add regression coverage through `_unbind_plugin()` (including a partial wrapping a bound method).

Regarding the disable-path discussion: I checked #9916 and understand that activation gating is deliberately handled separately. Keeping registrations on disable is consistent with that split, so I would not require disable-time unregistration here. Please update this PR's description, which currently says `_unbind_plugin()` is the shared cleanup path for disable as well, and link #9916 to make the scope clear. The ownership resolver in #9916 should also account for partials; it currently uses the same module/owner checks.

Validation: the existing plugin-manager/dashboard suites passed locally (`236 passed`), and Ruff checks passed for all three changed files. The additional partial-handler cleanup regression failed on this head as described above.

PR #9956, opened 2026-09-05
fix(command): keep an argument that equals another alias of the same command

he-yufeng:
Fixes a silent argument-parsing bug in `CommandFilter`. When a command has an alias and the user's first argument happens to equal that alias, the argument is silently dropped.

`CommandFilter.filter` loops over `get_complete_command_names()` (the command name plus every alias) and strips the matched prefix from `message_str`, but it never `break`s after the first match. So once the command prefix is stripped, the loop keeps testing the remaining text against the other names, and if the first argument equals one of the command's own aliases it gets stripped a second time. No error, no log, the argument just disappears.

## What changed

- `astrbot/core/star/filter/command.py`: break out of the name loop after the first match, and try the longest name first. The name set is built from a `set`, so its iteration order is not deterministic; sorting longest-first keeps overlapping names (an alias that extends the command name, like `show` vs `show all`) resolving to the most specific one.
- `tests/test_command_filter.py`: regression tests for an argument colliding with an alias (with and without trailing text), a normal argument, and prefix-overlapping names.

## Verification

`pytest tests/test_command_filter.py` on Python 3.12: 6 passed with the fix. Against master the three bug-pinning tests fail (the alias argument is swallowed; the overlapping name leaks `all` into the argument) and the three behavior tests still pass. ruff 0.15.22 check and format clean on both files.

This replaces #9294, same fix and tests; that branch picked up unrelated junk files from a sync artifact and its diff no longer renders on GitHub.

## Summary by Sourcery

Fix command parsing so matched command prefixes are removed only once and arguments are preserved reliably.

Bug Fixes:
- Preserve command arguments that match another alias of the same command during parsing.
- Resolve overlapping command names by selecting the most specific match.

Tests:
- Add regression coverage for alias-colliding arguments, normal arguments, and overlapping command names.

### DIFF
--- a/astrbot/core/star/filter/command.py
+++ b/astrbot/core/star/filter/command.py
@@ -198,10 +198,18 @@ class CommandFilter(HandlerFilter):
         # 检查是否以指令开头
         message_str = re.sub(r"\s+", " ", event.get_message_str().strip())
         ok = False
-        for full_cmd in self.get_complete_command_names():
+        # Try the longest name first so that when several complete command names
+        # share a prefix (e.g. an alias "show" and "show all"), the most specific
+        # one wins and the match is stable. get_complete_command_names() is built
+        # from a set, so its own order is not deterministic; sort here rather than
+        # there because other callers rely on [0] being the primary command name.
+        for full_cmd in sorted(
+            self.get_complete_command_names(), key=lambda name: (-len(name), name)
+        ):
             if message_str.startswith(f"{full_cmd} ") or message_str == full_cmd:
                 ok = True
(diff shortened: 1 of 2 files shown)


anka-afk (approved):
这个很好 @Soulter

he-yufeng:
@anka-afk 谢谢评审。CI 是绿的、main 没有冲突，这个修复现在可以直接合。顺带在 master 上又核实了一遍：delete_kb_by_id 仍然只删 KnowledgeBase 行，文档和多媒体记录还是孤儿，这个补丁的行为没有变化。

PR #9976, opened 2026-09-08
fix(star): bind and unload hooks registered from plugin submodules

### LINKED ISSUE #9938: [Bug] on_decorating_result() 报错缺失event参数
### 问题描述

在使用钩子时
@filter.on_decorating_result()
    async def on_decorating_result(self, event: AstrMessageEvent):
    .......
报错：
[21:48:44.511] [Core] [ERRO] [v4.27.4] [result_decorate.stage:182]: Traceback (most recent call last):
  File "/home/xie/AstrBot/astrbot/core/pipeline/result_decorate/stage.py", line 173, in process
    await handler.handler(event)
          ^^^^^^^^^^^^^^^^^^^^^^
TypeError: AstrBot_Plugin_tts_vllm_omni.on_decorating_result() missing 1 required positional argument: 'event'
但使用方法与文档一致且不太可能是缩进问题（self，在插件类下）
(照片因网络问题无法上传)

### 如何复现？

eg.
```python

    @filter.on_decorating_result()
    async def on_decorating_result(self, event: AstrMessageEvent):
        """结果装饰钩子：按概率将纯文本回复替换为语音。"""
        if not self._if_random_tts or self._random_tts is None:
            return
        chain = event.get_result().chain
        event.get_result().chain = await self._random_tts.maybe_replace(
            chain, self._random_tts_factor,
        )

```

### AstrBot 版本

 v4.27.4

### 操作系统

Linux

### 部署方式

shell(uv)

### 使用的消息平台适配器

OneBot V11

### 错误日志

```text
[21:48:47.923] [Core] [ERRO] [v4.27.4] [result_decorate.stage:182]: Traceback (most recent call last):
  File "/home/xie/AstrBot/astrbot/core/pipeline/result_decorate/stage.py", line 173, in process
    await handler.handler(event)
          ^^^^^^^^^^^^^^^^^^^^^^
TypeError: AstrBot_Plugin_tts_vllm_omni.on_decorating_result() missing 1 required positional argument: 'event'
```

### 辅助信息

_No response_

### 检查清单

- [ ] (可选)我愿意提交 PR 以帮助修复该问题。

he-yufeng:
### Motivation / 动机

Fixes #9938. A hook defined in a helper submodule of a plugin package (e.g. `my_plugin/hooks.py` imported from the entry module) was never bound to the plugin instance, so the pipeline called the raw function and raised `TypeError: on_decorating_result() missing 1 required positional argument: 'event'`.

插件的钩子如果写在辅助子模块里（而不是入口主模块），加载时不会被绑定到插件实例，管线调用裸函数直接报缺参。

### Modifications / 改动点

- `astrbot/core/star/star_handler.py`: `get_handlers_by_module_name` now matches the whole plugin package (entry module plus anything under it). Handler registration records `handler.__module__` (where the hook is defined), but both the load/`turn_on_plugin` binding loop and `_unbind_plugin` matched on the exact entry module only, so submodule hooks were never bound and were also leaked in the registry at unload.
- `tests/test_plugin_manager.py`: new regression test.

### Test Results / 测试结果

New test `test_submodule_hook_binds_on_activation_and_unloads_with_plugin`: registers an `on_decorating_result` hook under `<plugin>.main.hooks`, loads the plugin deactivated twice (handler stays raw), activates it and asserts the handler is a `functools.partial` bound to the live instance and callable with just the event, then unbinds and asserts the handler leaves the registry.

- Red on unmodified master (fails at the binding assertion), green here.
- Full `tests/test_plugin_manager.py`: 63 passed.
- `ruff format` + `ruff check` clean on both touched files.

Environment: macOS arm64, Python 3.12 via `uv run pytest`.

---

### Checklist / 检查清单

- [x] 😊 This is a bug fix for an existing issue (#9938); no new feature. / 这是针对已有 issue 的修复，无新功能。
- [x] 👀 My changes have been well-tested, with verification steps and results above. / 改动已充分测试，验证步骤与结果见上。
- [x] 🤓 No new dependencies introduced. / 未引入新依赖。

## Summary by Sourcery

Fix plugin lifecycle handling so hooks registered from helper submodules are bound, dispatched, and unloaded with their owning plugin.

Bug Fixes:
- Bind hooks defined in plugin submodules to the plugin instance during activation and remove them correctly when the plugin is unloaded.
- Resolve plugin ownership for submodule handlers so activation filtering and event dispatch select them correctly.

Tests:
- Add regression coverage confirming submodule hooks remain raw while inactive, bind on activation, dispatch correctly, and are removed on unload.

### DIFF
--- a/astrbot/core/star/star_handler.py
+++ b/astrbot/core/star/star_handler.py
@@ -6,9 +6,38 @@ from dataclasses import dataclass, field
 from typing import Any, Generic, Literal, TypeVar, overload
 
 from .filter import HandlerFilter
-from .star import star_map
+from .star import StarMetadata, star_map
 
 T = TypeVar("T", bound="StarHandlerMetadata")
+
+
+def plugin_package_root(plugin: StarMetadata) -> str | None:
+    """The module subtree a plugin owns, derived from its root directory.
+
+    ``module_path`` points at the entry module (``...my_plugin.main``), but a
+    plugin may define hooks in sibling modules (``...my_plugin.hooks``), so
+    matching needs the package root (``...my_plugin``).
+    """
+    if not plugin.root_dir_name:
+        return None
+    base = "astrbot.builtin_stars." if plugin.reserved else "data.plugins."
+    return base + plugin.root_dir_name
+
+
+def plugin_for_handler_module(module_path: str) -> StarMetadata | None:
+    """Resolve the plugin a handler module belongs to.
+
+    ``star_map`` is keyed by entry module, so a helper-module handler misses
+    the exact lookup; fall back to matching the plugin's package root.
+    """
+    plugin = star_map.get(module_path)
+    if plugin is not None:
+        return plugin
+    for plugin in star_map.values():
+        root = plugin_package_root(plugin)
+        if root and (module_path == root or module_path.startswith(root + ".")):
+            return plugin
+    return None
 
 
 class StarHandlerRegistry(Generic[T]):
@@ -162,12 +191,12 @@ class StarHandlerRegistry(Generic[T]):
                 continue
             # 过滤启用状态
             if only_activated:
-                plugin = star_map.get(handler.handler_module_path)
+                plugin = plugin_for_handler_module(handler.handler_module_path)
                 if not (plugin and plugin.activated):
                     continue
             # 过滤插件白名单
             if plugins_name is not None and plugins_name != ["*"]:
-                plugin = star_map.get(handler.handler_module_path)
+                plugin = plugin_for_handler_module(handler.handler_module_path)
                 if not plugin:
                     continue
                 if (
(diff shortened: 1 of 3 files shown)


he-yufeng:
Both findings were real, thanks. Fixed at 0f810a01.

1. The binding predicate indeed only covered the entry module and its children, so a sibling helper module (`my_plugin/hooks.py` next to `my_plugin/main.py`, which is the actual #9938 layout) still missed. Callers now pass the plugin's package root derived from `root_dir_name` (via `plugin_package_root`), and `get_handlers_by_module_name` matches the whole package subtree.

2. `star_map` is keyed by the entry module, so a bound submodule handler was still filtered out by `get_handlers_by_event_type`. The dispatch lookup now goes through `plugin_for_handler_module`, which falls back to the package-root match when the exact entry lookup misses.

The regression test now uses the sibling layout (the previous version only covered a child of the entry module, which the old predicate already matched, so it could not catch either gap) and additionally asserts the bound handler is selected by `get_handlers_by_event_type`, exercising the dispatch path instead of only the binding. Verified red-green both ways: the updated test fails on the previous predicate and dispatch lookup, passes on this head; full `test_plugin_manager.py` 63/63 green.

PR #9985, opened 2026-09-08
fix: make TTS voice replies work when streaming output is enabled

foorgange:
## Problem

When both "streaming output" (`provider_settings.streaming_response`) and TTS voice replies are enabled (a `text_to_speech` provider configured, `provider_tts_settings.enable = true`), replies are **never** converted into voice messages. Users only ever receive streamed text, and `provider_tts_settings.trigger_probability` has no effect in this mode. The only workaround today is to disable streaming output globally, which is not documented and is invisible to most users.

### Why this happens

Text-to-voice conversion lives in exactly one place: `ResultDecorateStage` (see the "TTS" block in `result_decorate/stage.py`). That stage only processes fully generated, non-streaming results:

- `STREAMING_RESULT` results return early from the decorate stage and are delivered directly to the platform adapter by the respond stage.
- `STREAMING_FINISH` results are also skipped ("streaming output does not execute the following logic").

At the same time, the agent sub-stages (`InternalAgentSubStage` for the local tool-loop runner, `ThirdPartyAgentSubStage` for Dify/Coze/DashScope/DeerFlow runners) decide whether a reply streams purely from `provider_settings.streaming_response` and per-message `enable_streaming` flags. They never look at `provider_tts_settings`, so a reply that was meant to be voiced is silently streamed instead and bypasses the only TTS code path.

## Fix

Keep the probability decision in one place per reply path and make streaming respect it:

1. In `InternalAgentSubStage` and `ThirdPartyAgentSubStage`, right where the agent knows the reply will actually be streamed (`streaming_response and not stream_to_general`), roll the dice once. When `provider_tts_settings.enable` is on and the roll hits `trigger_probability`, and a usable TTS provider is configured for the session, mark the event with a `tts_forced` extra and switch the reply to non-streaming. The complete result then flows through `ResultDecorateStage` and is converted into a voice message.
   - In `InternalAgentSubStage` the roll happens **before** the agent runner is built and reset, so the runner itself is created without streaming; otherwise the runner still emits streaming deltas and the non-streaming path never produces the final result the decorate stage needs.
   - Live mode is excluded: it runs its own streaming TTS pipeline (`run_live_agent`) and must not be diverted.
   - The probability roll uses a strict comparison (`random.random() < probability`), so `trigger_probability = 0` can never force a voice reply.
   - The dice only forces non-streaming when a usable TTS provider actually exists for the session, so a misconfigured TTS setup cannot silently disable streaming for regular chat replies.
2. In `ResultDecorateStage`, consume the probability only once: when `tts_forced` is present, the dice was already rolled upstream, so the probability for that reply is `1.0`; ordinary non-streaming results keep using the configured `trigger_probability` as before.

Resulting behavior for any configuration that enables TTS:

| Streaming output | Probability roll | Behavior |
| --- | --- | --- |
| enabled | hit | reply is fully generated first, then sent as a TTS voice message |
| enabled | miss | reply is streamed as text (previous behavior) |
| disabled | hit | reply is sent as a TTS voice message (previous behavior) |
| disabled | miss | reply is sent as plain full text (previous behavior) |

Setting `trigger_probability` to `0` restores the previous behavior in every mode. When no TTS provider is enabled, nothing changes at all.

## Manual testing

Verified on a deployment running v4.26.8 (AstrBot docker image) with a QQ (aiocqhttp/NapCat) bot:

- TTS provider + streaming output enabled, `trigger_probability = 0.99`: roughly 99% of replies arrive as voice messages (logs show `TTS request` / `TTS result` only for those replies), the rest stream as text.
- `trigger_probability = 0.5`: voice and streamed text replies are interleaved at the expected ratio.
- `trigger_probability = 0`: all replies stream again, identical to the previous behavior.
- Streaming output disabled: voice replies follow `trigger_probability` exactly, unchanged from before this fix.
- Both agent families covered: the local runner path (`InternalAgentSubStage`, QQ group chat with tool-loop agent) and the third-party runner path (`ThirdPartyAgentSubStage`).

Code quality: `python -m ruff format` and `python -m ruff check` pass on all modified files (ruff 0.15.x, matching the project pin).

## Notes

- Comments added in English, following the project convention.
- No new configuration keys and no default value changes: `trigger_probability` already defaults to `1.0`, which now simply means "always voice" for streaming users instead of "never voice".
- `tts_forced` is only ever set when a stream would actually have happened for that reply; it has no side effects on non-streaming platforms or when TTS is disabled.
- This pull request supersedes #9982 (closed), which contained an earlier revision of the same change where the dice roll happened after the internal agent runner was built.

## Summary by Sourcery

Fix streamed agent replies so configured TTS responses can be delivered as voice messages without changing ordinary streaming behavior.

Bug Fixes:
- Allow TTS voice replies to work when streaming output is enabled by selectively switching eligible streamed responses to complete generation before voice conversion.

Enhancements:
- Coordinate TTS probability decisions across agent stages and result decoration so each reply is evaluated only once while preserving existing behavior for non-streaming replies, live mode, and unavailable TTS providers.

### DIFF
--- a/astrbot/core/pipeline/process_stage/method/agent_sub_stages/internal.py
+++ b/astrbot/core/pipeline/process_stage/method/agent_sub_stages/internal.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import base64
+import random
 from collections.abc import AsyncGenerator
 from dataclasses import replace
 
@@ -225,6 +226,39 @@ class InternalAgentSubStage(Stage):
             if await call_event_hook(event, EventType.OnWaitingLLMRequestEvent):
                 return
 
+            stream_to_general = (
+                self.unsupported_streaming_strategy == "turn_off"
+                and not event.platform_meta.support_streaming_message
+            )
+
+            # When the reply would be streamed but TTS voice replies are
+            # enabled, streaming delivery bypasses the result decorate stage
+            # (the only place text is converted into voice messages). Roll the
+            # dice before the agent runner is built and reset, so forced
+            # replies are created without streaming. Live mode has its own
+            # streaming TTS pipeline and is excluded.
+            if (
+                streaming_response
+                and not stream_to_general
+                and event.get_extra("action_type") != "live"
+            ):
+                tts_cfg = self.ctx.astrbot_config.get("provider_tts_settings", {})
+                if tts_cfg.get("enable"):
+                    try:
+                        tts_prob = float(tts_cfg.get("trigger_probability", 1.0))
+                    except (TypeError, ValueError):
+                        tts_prob = 1.0
+                    if random.random() < max(0.0, min(tts_prob, 1.0)):
+                        # Only force when a usable TTS provider exists, so a
+                        # misconfigured TTS setup cannot silently disable
+                        # streaming for regular chat replies.
+                        tts_provider = await self.ctx.plugin_manager.context.get_using_tts_provider_async(
+                            event.unified_msg_origin
+                        )
+                        if tts_provider is not None:
+                            event.set_extra("tts_forced", True)
+                            streaming_response = False
+
             async with session_lock_manager.acquire_lock(event.unified_msg_origin):
                 logger.debug("acquired session lock for llm request")
                 agent_runner: AgentRunner | None = None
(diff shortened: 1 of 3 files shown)


foorgange:
Thanks for the review. The provider lookup in the dice roll is now wrapped: a session that resolves to a provider of the wrong type raises ValueError from \get_using_tts_provider_async\, which is caught and treated as \	ts_provider = None\, so the reply keeps the normal streaming path instead of aborting. Applied in both \InternalAgentSubStage\ and \ThirdPartyAgentSubStage\ (commit 096273c).

foorgange:
Superseded by a new pull request on a clean branch containing the addressed provider-lookup fix; the review bot evaluates each pull request once.

PR #9739, opened 2026-08-19
feat: add permission check for /new command

Rain-0x01-39:
`/new` 命令缺少权限控制，在群聊且会话隔离关闭的场景下（共享会话），任意成员都可以创建新会话并切换，影响其他用户的对话上下文。而 `/reset` 已有对应的权限检查，`/new` 与之不一致。

### Modifications / 改动点

- `astrbot/builtin_stars/builtin_commands/commands/conversation.py`：为 `new_conv` 方法新增场景化权限检查，逻辑与 `reset` 一致——群聊+会话隔离关闭时默认要求 `admin`，其余场景 `member` 即可。权限可通过 Dashboard 的指令管理覆盖。

- [x] This is NOT a breaking change. / 这不是一个破坏性变更。

### Screenshots or Test Results / 运行截图或测试结果

---

### Checklist / 检查清单

- [ ] 😊 If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc. 
  / 如果 PR 中有新加入的功能，已经通过 Issue / 邮件等方式和作者讨论过。

- [ ] 👀 My changes have been well-tested, **and "Verification Steps" and "Screenshots" have been provided above**.
  / 我的更改经过了良好的测试，**并已在上方提供了“验证步骤”和“运行截图”**。

- [x] 🤓 I have ensured that no new dependencies are introduced, OR if new dependencies are introduced, they have been added to the appropriate locations in `requirements.txt` and `pyproject.toml`.
  / 我确保没有引入新依赖库，或者引入了新依赖库的同时将其添加到 `requirements.txt` 和 `pyproject.toml` 文件相应位置。

- [x] 😮 My changes do not introduce malicious code.
  / 我的更改没有引入恶意代码。

## Summary by Sourcery

Enforce context-aware permissions for creating new conversations.

Bug Fixes:
- Restrict the `/new` command in shared group conversations to administrators, preventing members from changing the shared conversation context.

Enhancements:
- Align `/new` permission behavior with `/reset` while preserving configurable per-scenario permission overrides.

PR #10004, opened 2026-09-09
feat: unify /new and /reset behavior

w31r4:
Both `/new` and `/reset` feel like starting over, but previously they used different implementations: `/new` created a conversation, while `/reset` cleared the current conversation's stored history. This PR unifies their restart behavior and manages their permissions through command management.

### Modifications

- For the built-in Agent, both commands stop running tasks in the current session, create and select an empty conversation, and carry over the current persona selection. The old conversation and its history remain in the database. Only the restarted session's group-context cache is cleared.
- Keep both command entries and their individual permission, enable/disable, and rename settings.
- Default both commands to **Follow Conversation Isolation**: everyone can use them in private chats and actually isolated group conversations; shared group conversations require AstrBot administrator permission. Unsupported platforms retain the shared-group restriction even if the isolation switch is enabled.
- Manage each command under **Extensions → Handlers → Command**, with **Show System Plugin Commands** selected. The choices are **Everyone**, **Administrators Only**, **Administrators Only in Group Chats**, and **Follow Conversation Isolation**. The first three remain fixed when isolation changes; explicitly saved permissions are preserved. **Administrators Only** also restricts private chats. Administrators are configured AstrBot administrator IDs, not automatically detected group administrators.
- A command's selected policy applies across configuration profiles. The isolation-following policy uses the incoming message's actual isolation state, so different profiles do not overwrite each other's permissions. **Isolate Conversation** remains a separate configuration setting; no additional platform permission switch is introduced.
- Retain the existing third-party Agent restart flow: clear the saved external conversation ID so the next message starts a new conversation. DeerFlow also attempts to delete the old remote thread; history retention is not guaranteed for third-party services.
- Update Chinese, English, Japanese, and Russian UI text, plus Chinese and English user/developer documentation. Add regression checks that documentation uses the current WebUI labels.

- [ ] This is NOT a breaking change.

Compatibility notes: `/reset` creates a new conversation instead of clearing the current one. Without an explicit permission override, ordinary group members can use `/new` and `/reset` only when their conversations are actually isolated. Existing command-management permissions and isolation settings are preserved. Legacy `group_unique_on`, `group_unique_off`, and `private` values under `alter_cmd.astrbot.reset` are no longer read; users who customized these values should select the desired permission in command management. Selecting **Follow Conversation Isolation** restores automatic checks.

Local validation on WSL/Linux for `1a2492a09`:

- Full pytest suite in an isolated checkout of merge snapshot `748e515ec03b822b372ee3d52c59218418e9d4e3` with this change applied: **2600 passed**. The final Russian checkbox wording correction was subsequently covered by the translation checks and dashboard build.
- Targeted restart, permission, pipeline, API, and registry-isolation regression tests: **169 passed**. Coverage includes profile interleaving, isolation toggles, unsupported platforms, explicit overrides, permission persistence/reload, history preservation, and both command entries.
- `ruff format .` and `ruff check .`: passed. Changed test files were also formatted and checked explicitly with `--no-respect-gitignore`.
- `cd dashboard && pnpm build`: passed, including TypeScript checks.
- Chromium checks in all four UI languages: permission labels and hints, all four choices, update payloads, independent command settings, refresh, filtering, and details. These checks use mocked API responses, not a live backend E2E setup.

Verification: run the pytest suite using the CI repository/data-directory layout, then run Ruff and the dashboard build. In command management, select **Follow Conversation Isolation** for `new` and `reset`. On a platform supporting isolation, verify that a regular group member is denied with isolation disabled and allowed with it enabled; disable isolation again and verify access is restricted. A manually selected fixed permission must remain unchanged.

No real QQ/NapCat connection, live model invocation, or external Agent service E2E was used for this revision. Windows and macOS validation is left to CI. No screenshots are attached.

---

### Checklist

- [x] The feature and permission behavior have been discussed with the maintainers.
- [x] Verification steps and test results are provided above.
- [x] No new project dependencies are introduced; Playwright is used only for temporary local browser checks.
- [x] The changes do not introduce malicious code.


astrbot/core/config/default.py
```diff
@@ -4243,10 +4247,15 @@
...
                     "platform_settings.unique_session": {
-                        "description": "隔离会话",
+                        "description": "隔离对话",
                         "type": "bool",
                         "hint": "启用后，群成员的上下文独立。",
                     },
+                    "platform_settings.allow_member_new_conversation": {
+                        "description": "允许非管理员在群聊中新建对话",
+                        "type": "bool",
+                        "hint": "允许普通成员使用 /new 和 /reset。共享上下文时会切换整个群的当前对话；成员隔离时只影响自己的对话。内置 Agent 的旧记录保留。命令管理中的权限限制仍然有效。",
```
Soulter: nit: 允许普通成员使用 /new 和 /reset 指令（关闭隔离对话时指令会切换整个群的对话；开启隔离对话时指令只影响自己的对话。不会删除旧对话）。命令管理中的权限限制仍然有效。

顺便可以改一下 i18n 的语料

astrbot/core/config/default.py
RC-CHN, line 1: 这一部分的交互是否考虑到放在extension/components内进行，不要耦合到系统配置项上做特殊处理

Image

w31r4: 正在迁移

Soulter (approved):

Rain-0x01-39:
`/new` 命令现在还是没有任何权限检查吗？我的 #9739 解决了这个问题，未合。

PR #10008, opened 2026-09-09
fix(cron): deliver proactive agent final text to the bound session

xiaoyuyu6420:
## Problem

Proactive agent runs (cron `active_agent` jobs and background-task result wakeups) discard every `AgentResponse`: both call sites consume `runner.step_until_done()` with `async for _ ... pass`. The final assistant text is only folded into the persisted history summary — it is never sent to the delivery session.

Two details make this worse and turn it into the exact symptom reported in #9980 ("only intermediate messages, no final result"):

1. At forced wrap-up (`ToolLoopAgentRunner.step_until_done`, max steps reached) all tools are removed before the final step, so the model cannot use `send_message_to_user` to deliver its forced summary.
2. `SendMessageToUserTool`'s description says *"For other normal text replies, you can output directly and no need to use this tool"* — models that follow that hint output plain text, which is exactly the text being discarded.

This is the remaining gap from the #9980 breakdown; #9987 fixed the ERROR-state propagation and #9992 fixed compression-config sharing. This PR fixes final-response delivery for both wake paths.

## Fix

- `CronJobManager._woke_main_agent()` and `AstrAgentToolExec._wake_main_agent_for_background_result()` now send the final assistant text through the cron event after `persist_agent_history()`, when:
  - the runner finished and `llm_resp.role == "assistant"` with non-empty text (error responses are never pushed to chats);
  - the job has a real delivery session (cron jobs without one keep their current behavior);
  - the model did not already deliver this exact text to the current session via `send_message_to_user` earlier in the same run.
- The double-delivery guard reads the per-run record that `SendMessageToUserTool` already writes to the event; the key is now a shared constant (`SENT_TO_CURRENT_SESSION_PLAIN_TEXTS_EXTRA_KEY`).
- A failed delivery is logged and does **not** fail the job — the agent work and history persistence already succeeded.

## Testing

- New tests in `tests/unit/test_cron_manager.py` (`TestWokeMainAgentFinalDelivery`):
  - final assistant text is delivered to the delivery session;
  - no duplicate send when the tool already delivered the same text;
  - no send when there is no delivery session / empty text / non-assistant role;
  - a delivery failure is swallowed and does not mark the job failed.
- Full suite: 2420 passed.

Fixes part of #9980 (final-response delivery half).

## Summary by Sourcery

Deliver final assistant text from forced proactive-agent wrap-ups to their bound sessions without duplicating messages or affecting job success.

New Features:
- Deliver forced-wrap-up assistant responses from cron jobs and background-task wakeups to their bound sessions.

Bug Fixes:
- Prevent proactive agent runs from discarding final assistant text when tool execution reaches the maximum step limit.
- Avoid duplicate final-message delivery when the same text was already sent through send_message_to_user.
- Ensure delivery failures are logged without failing the completed agent job.

Enhancements:
- Expose whether an agent run ended through forced max-step completion and centralize tracking of texts sent to the current session.

Tests:
- Add coverage for forced-wrap-up delivery, duplicate suppression, invalid delivery conditions, session targeting, normal-completion silence, delivery failures, and max-step state tracking.

### DIFF
--- a/astrbot/core/agent/runners/tool_loop_agent_runner.py
+++ b/astrbot/core/agent/runners/tool_loop_agent_runner.py
@@ -127,6 +127,8 @@ class ToolLoopAgentRunner(BaseAgentRunner[TContext]):
         "Stop calling tools, and based on the information you have gathered, "
         "summarize your task and findings, and reply to the user directly."
     )
+    # Class-level default so the read-only property is safe before reset().
+    _reached_max_steps: bool = False
     SKILLS_LIKE_REQUERY_INSTRUCTION_TEMPLATE = (
         "You have decided to call tool(s): {tool_names}. Now call the tool(s) "
         "with required arguments using the tool schema, and follow the existing "
--- a/astrbot/core/agent/runners/tool_loop_agent_runner.py
+++ b/astrbot/core/agent/runners/tool_loop_agent_runner.py
@@ -233,6 +233,7 @@ class ToolLoopAgentRunner(BaseAgentRunner[TContext]):
     ) -> None:
         self.req = request
         self.streaming = streaming
+        self._reached_max_steps = False
         self.enforce_max_turns = enforce_max_turns
         self.llm_compress_instruction = llm_compress_instruction
         self.llm_compress_keep_recent_ratio = llm_compress_keep_recent_ratio
@@ -1068,6 +1069,16 @@ class ToolLoopAgentRunner(BaseAgentRunner[TContext]):
 
             self.req.append_tool_calls_result(tool_calls_result)
 
+    @property
+    def reached_max_steps(self) -> bool:
+        """Whether this run ended via the forced final response at max steps.
+
+        Only meaningful after ``step_until_done`` has run; a forced wrap-up
+        removes all tools, so any final text the model produced there could
+        not be delivered through ``send_message_to_user``.
+        """
+        return self._reached_max_steps
+
     async def step_until_done(
         self, max_step: int
     ) -> T.AsyncGenerator[AgentResponse, None]:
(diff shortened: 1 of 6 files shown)


xiaoyuyu6420:
Addressed the Sourcery bug_risk in a11ce12f1: the cron path now delivers via `Context.send_message(delivery_session_str, ...)` instead of `cron_event.send(...)`, so the gate condition and the delivery target are the same variable — a caller passing mismatched `session_str`/`delivery_session_str` can no longer route the text to the wrong session (no-platform and exceptions are logged without failing the job). The background-result path still sends through its cron event, whose session is built from the original event's own `unified_msg_origin`, so target and gate are identical there by construction.

Added `test_delivery_targets_delivery_session_not_event_session` covering the mismatch case.

kilisamemarisaaa (approved):
Reviewed the current head and the forced-wrap-up delivery path. The 
eached_max_steps guard is set only where tools are removed, while normal completion remains silent; the cron path now targets delivery_session_str directly and the duplicate-text check is scoped to the current event. The added tests cover target mismatch, duplicate suppression, invalid responses, delivery failure, and the runner flag, and the full CI matrix is green. I found no blocking issue in the current diff.

xiaoyuyu6420:
Closing this to avoid a dangling PR: Soulter closed #9980 as completed on Sep 12.

I checked current master to confirm whether this is still needed before closing:

- The final-response delivery gap for cron / background-task wakeups is **not** covered by any merged work — #10025 (skills-like streaming bridge), #10028 (session model fallback) and the earlier #9987 / #9992 all touch different paths.
- This PR adds `reached_max_steps` on `ToolLoopAgentRunner` and delivers the forced-wrap-up final text to the delivery session on both wake paths, with tests (2424 passing at last push).

Since the issue is closed and there has been no maintainer review, the honest state is that this half is currently not in scope. The branch `fix/9980-proactive-agent-final-delivery` and its commits remain; happy to reopen or rebase if the maintainers decide the delivery half should be picked up.

PR #10042, opened 2026-09-11
fix: preserve image attachment paths after event cleanup

Soulter:
Image attachment text currently exposes compressed image paths that are deleted when the event finishes. Image preprocessing also deletes the image files needed by later tasks. Preserve the preprocessed image and reference its path in the model request so subsequent tool calls can reuse it after event cleanup.

### Modifications

- For current and quoted images, register the original temporary file for deletion only after JPEG conversion returns a different path. Keep the converted output, unchanged images, and originals whose conversion fails.
- Put the pre-compression image path in attachment text while continuing to send compressed images as visual input and clean up those compressed copies when the event finishes.
- Remove Mattermost's attachment-path metadata and blanket event cleanup registration. Its attachments still follow the common image and audio preprocessing rules.

Retained files remain in the existing storage locations and are still subject to temporary-directory capacity cleanup and manual cache cleanup; this change does not provide permanent storage.

- [x] This is not a breaking change.

- `uv run pytest tests/test_preprocess_stage.py tests/test_mattermost_adapter.py tests/unit/test_astr_main_agent.py -q` — 140 passed.
- Tests cover current and quoted images, successful and failed conversion, unchanged formats, compression enabled and disabled, Mattermost attachments, and file survival after event cleanup.
- `ruff format .`, `ruff check .`, and `git diff --check` — passed.

### Checklist

- [x] Changes have been tested, with verification commands and results provided above.
- [x] Changes do not introduce malicious code.

## Summary by Sourcery

Preserve image attachment files needed by later model and tool calls after event cleanup.

Bug Fixes:
- Preserve usable original image paths in attachment text after event cleanup while continuing to use and clean up compressed visual copies.

Enhancements:
- Align image preprocessing cleanup with whether JPEG conversion creates a replacement file, including quoted and Mattermost attachments.
- Route Mattermost attachment files through the shared media preprocessing and cleanup behavior instead of blanket event-level cleanup.

Tests:
- Add coverage for current and quoted images, conversion outcomes, compression settings, Mattermost attachments, and file survival after cleanup.

### DIFF
--- a/astrbot/core/pipeline/preprocess_stage/stage.py
+++ b/astrbot/core/pipeline/preprocess_stage/stage.py
@@ -131,9 +131,9 @@ class PreProcessStage(Stage):
             elif isinstance(component, Image):
                 try:
                     original_path = await component.convert_to_file_path()
-                    self._track_temp_media(event, original_path)
                     image_path = await ensure_jpeg(original_path)
-                    self._track_temp_media(event, image_path)
+                    if image_path != original_path:
+                        self._track_temp_media(event, original_path)
                     component.file = image_path
                     component.path = image_path
                     # Image.convert_to_file_path() prefers url, so keep it aligned.
@@ -167,9 +167,9 @@ class PreProcessStage(Stage):
                     elif isinstance(reply_comp, Image):
                         try:
                             original_path = await reply_comp.convert_to_file_path()
-                            self._track_temp_media(event, original_path)
                             image_path = await ensure_jpeg(original_path)
-                            self._track_temp_media(event, image_path)
+                            if image_path != original_path:
+                                self._track_temp_media(event, original_path)
                             reply_comp.file = image_path
                             reply_comp.path = image_path
                             # Image.convert_to_file_path() prefers url, so keep it aligned.
--- a/astrbot/core/pipeline/preprocess_stage/stage.py
+++ b/astrbot/core/pipeline/preprocess_stage/stage.py
@@ -131,10 +131,6 @@ class PreProcessStage(Stage):
             elif isinstance(component, Image):
                 try:
                     original_path = await component.convert_to_file_path()
-                    # Keep the local reference even if JPEG conversion fails.
-                    component.file = original_path
-                    component.path = original_path
-                    component.url = original_path
                     image_path = await ensure_jpeg(original_path)
                     if image_path != original_path:
                         self._track_temp_media(event, original_path)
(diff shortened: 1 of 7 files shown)


astrbot/core/pipeline/preprocess_stage/stage.py
```diff
@@ -131,9 +131,9 @@ async def process(
             elif isinstance(component, Image):
                 try:
                     original_path = await component.convert_to_file_path()
-                    self._track_temp_media(event, original_path)
                     image_path = await ensure_jpeg(original_path)
```
kilisamemarisaaa, line 134: Please preserve cleanup ownership when JPEG normalization fails after materializing a remote/data image. I reproduced this at `69cd1d8` with an invalid image data URI: `convert_to_file_path()` created `media_image_*.png`, then `ensure_jpeg()` raised before either path was tracked. The catch left the component pointing at the data URI, `_temporary_local_files` was empty, and event cleanup left the new file on disk. The base revision tracks that materialized path and removes it. Please track the materialized original on this exception path (or otherwise delete it when it is no longer the component reference); the same issue exists in the quoted-image branch.

astrbot/core/platform/sources/mattermost/mattermost_adapter.py
```diff
@@ -230,12 +230,10 @@ async def convert_message(
             )
 
         if file_ids:
-            (
-                attachment_components,
-                temp_paths,
-            ) = await self.client.parse_post_attachments(file_ids)
+            attachment_components, _ = await self.client.parse_post_attachments(
```
kilisamemarisaaa, line 233: Discarding `temp_paths` leaks the downloaded source for converted Mattermost audio. `parse_post_attachments()` writes (for example) `mattermost_<id>.ogg`, converts it to a separate WAV, and returns a `Record` that only references the WAV. Preprocessing tracks and later deletes that WAV, but the original OGG is now unreachable and remains until the coarse temp-directory sweep. I reproduced this at `69cd1d8`: cleanup removed `converted.wav` but left the original OGG; the base revision removed both. The new test uses the same `voice.wav` as both the downloaded source and conversion result, so it cannot exercise this two-path case. Please retain cleanup ownership for source paths that are superseded during attachment parsing.

kilisamemarisaaa (changes requested):
I found two temp-file ownership regressions at this exact head. Both reproduce against the changed paths and are not covered by the otherwise-passing focused suites, so I am requesting changes until the materialized source files are cleaned up on these branches.

PR #10049, opened 2026-09-11
fix: keep inline Face components in the same bubble during segmented reply

### LINKED ISSUE #10047: [Bug] 开启分段回复（segmented_reply）时，消息链中的 Face（QQ表情）会被作为独立消息切碎单独发送
### 问题描述

当在 `platform_settings` 中启用了**分段回复（`segmented_reply.enable: true`）**时，若 LLM 返回的消息链中包含内嵌的 QQ 原生表情（`Face` 组件，例如在文字中间通过表情插件或正则将 `[狗头]` 转为 `Face(id=277)`），该表情无法与句子中的文字保持在同一个消息气泡中，而是会被作为一条独立的消息单独发送。

在 QQ 群聊中这会导致原本自然的一句话（如 `好的[狗头]，我知道了！`）被生硬切碎成 3 条消息：
1. `好的`
2. `[Face:277]`（独立发送一个黄豆表情）
3. `，我知道了！`

不仅影响阅读体验，还会触发单组件发送延迟造成断断续续刷屏。

#### 源码定位与分析
定位在核心响应管道文件 `astrbot/core/pipeline/respond/stage.py` 的 `RespondStage.process`：

```python
if self.is_seg_reply_required(event):
    header_comps = self._extract_comp(
        result.chain,
        {ComponentType.Reply, ComponentType.At},
        modify_raw_chain=True,
    )
    ...
    for comp in result.chain:  # 问题根源：简单遍历了 chain 里的每一个组件
        i = await self._calc_comp_interval(comp)
        await asyncio.sleep(i)
        try:
            if comp.type in need_separately:
                await event.send(result.derive([comp]))
            else:
                await event.send(result.derive([*header_comps, comp]))  # 每个非单独组件独立发送
                header_comps.clear()
```

分段回复的初衷应当是按句子/标点分段发送，但此处直接按组件 `for comp in result.chain:` 逐个发送。当链中存在 `Face` 或其他行内组件时，它们本该与前后文本同属一个分段气泡，却被强制拆分成单条消息。

#### 修复建议
在分段回复遍历前，将相邻的行内组件（如 `Plain` 和 `Face` 等非必须单独发送组件）打包合并为一个子消息链，按组进行分段发送：

```python
sub_chains = []
current_group = []

for comp in result.chain:
    if comp.type in need_separately:
        if current_group:
            sub_chains.append(current_group)
            current_group = []
        sub_chains.append([comp])
    else:
        # 行内组件（Plain, Face 等）合并在同一个消息气泡
        current_group.append(comp)

if current_group:
    sub_chains.append(current_group)

for group in sub_chains:
    i = await self._calc_comp_interval(group[0])
    await asyncio.sleep(i)
    try:
        if any(c.type in need_separately for c in group):
            await event.send(result.derive(group))
        else:
            await event.send(result.derive([*header_comps, *group]))
            header_comps.clear()
    except Exception as e:
        logger.error(f"Failed to send grouped message: {e}", exc_info=True)
```

### 如何复现？

1. 在全局设置 `cmd_config.json` 中开启分段回复（`platform_settings.segmented_reply.enable: true`）。
2. 在 LLM 回复链中构造或通过插件生成包含内嵌表情的消息链，例如：`result.chain = [Plain("好的"), Face(id=277), Plain("，我知道了！")]`。
3. 观察 AstrBot 发送到群聊的实际行为，会发现 `Face` 被拆解成一条单独的消息单独发出。

### AstrBot 版本

v4.27.5

### 操作系统

Linux

### 部署方式

Docker

### 使用的消息平台适配器

Onebot v11 (NapCat)

### 错误日志

```text
# 观察到 AstrBot 在分段回复时为每个组件单独调用一次 send：
[Core] [INFO] Prepare to send - User/123456: [Plain: 好的, Face: 277, Plain: ，我知道了！]
# 实际输出到 OneBot 的是连续的单元素调用：
# 1. send([Plain("好的")])
# 2. send([Face(277)]) -> 表现为独立的单黄豆表情消息气泡
# 3. send([Plain("，我知道了！")])
```

### 检查清单

- [x] (可选)我愿意提交 PR 以帮助修复该问题。

Shxiao101:
With `platform_settings.segmented_reply.enable: true`, the respond stage sends every component of the message chain as its own message. An inline QQ Face in the middle of text (e.g. `好的[Face:277]，我知道了！`) is therefore split into three separate bubbles: `好的` / `[Face:277]` / `，我知道了！`.

Fixes #10047

### Modifications / 改动点

- `astrbot/core/pipeline/respond/stage.py`: `_group_segment_chain()` now builds the segmented-reply bubbles: `Plain` and non-inline components start a new bubble, and a `Face` attaches to the preceding text bubble and glues the `Plain` following it into the same bubble, so an inline emoji never becomes (or creates) a bubble of its own. Adjacent `Plain` components without an emoji between them (segmentation-words output, #3959) keep their deliberate split.
- `_calc_comp_interval()` accepts a bubble (a list of components) and computes the log-method interval from the Plain word counts, counted per component and summed. Single-component behavior is unchanged.
- `tests/unit/test_respond_stage_segmented_reply.py`: regression tests for the glued sentence, leading Face, consecutive Plain split, Record staying alone, Face after media, glue stopping at media, and the interval calculation.

- `uv run pytest tests/unit` — 993 passed (10 deselected locally for an unrelated environment issue that also fails on clean master)
- `./scripts/pr_test_env.sh --profile neo` (repo-recommended local check set): `ruff format --check .` clean (504 files), `ruff check .` passed, Neo critical tests 13 passed, `main.py` startup smoke test on `http://localhost:6185` passed

---

### Checklist / 检查清单

- [ ] 😊 If there are new features added in the PR, I have discussed it with the authors through issues/emails, etc.

  / 我的更改经过了良好的测试，**并已在上方提供了"验证步骤"和"运行截图"**。

  / 我的更改没有引入恶意代码。

## Summary by Sourcery

Keep inline Face components with their surrounding text when sending segmented replies.

Bug Fixes:
- Keep inline Face components and their surrounding text in the same segmented-reply bubble instead of sending the emoji and following clause separately.

Enhancements:
- Preserve intentional splits between adjacent Plain components and keep non-inline components in separate bubbles while calculating reply intervals across all Plain text in a bubble.

Tests:
- Add regression coverage for inline and leading Faces, adjacent Plain splits, media boundaries, and grouped interval calculation.

### DIFF
--- a/tests/unit/test_respond_stage_segmented_reply.py
+++ b/tests/unit/test_respond_stage_segmented_reply.py
@@ -1,0 +1,56 @@
+"""Regression tests for segmented reply bubble grouping (#10047).
+
+With segmented reply enabled, every component used to be sent as its own
+message, so an inline Face in the middle of a sentence split the text into
+several bubbles. The stage now groups consecutive inline components
+(Plain / Face) so a sentence stays in one bubble.
+"""
+
+import math
+
+import pytest
+
+import astrbot.core.message.components as Comp
+from astrbot.core.pipeline.respond.stage import RespondStage
+
+
+def test_inline_face_stays_in_the_same_bubble_as_text():
+    stage = RespondStage()
+    chain = [
+        Comp.Plain(text="好的"),
+        Comp.Face(id=277),
+        Comp.Plain(text="，我知道了！"),
+    ]
+
+    segments = stage._group_segment_chain(chain)
+
+    assert len(segments) == 1
+    assert segments[0] == chain
+
+
+def test_components_that_need_separate_sending_stay_alone():
+    stage = RespondStage()
+    record = Comp.Record(file="file:///tmp/a.wav")
+    chain = [Comp.Plain(text="看这个"), record, Comp.Plain(text="好听吗")]
+
+    segments = stage._group_segment_chain(chain)
+
+    assert segments == [[chain[0]], [record], [chain[2]]]
+
+
+@pytest.mark.asyncio
+async def test_log_interval_for_a_bubble_uses_total_plain_word_count():
+    stage = RespondStage()
+    stage.interval_method = "log"
+    stage.log_base = 10.0
+
+    # "hello" + "world" -> 2 words -> log10(3) ~= 0.477
+    bubble = await stage._calc_comp_interval(
+        [Comp.Plain(text="hello"), Comp.Face(id=277), Comp.Plain(text="world")],
+    )
+    lower, upper = math.log(3, 10), math.log(3, 10) + 0.5
+    assert lower <= bubble <= upper
+
+    # A bubble without text keeps the non-Plain interval.
+    face_only = await stage._calc_comp_interval(Comp.Face(id=277))
+    assert 1 <= face_only <= 1.75
--- a/tests/unit/test_respond_stage_segmented_reply.py
+++ b/tests/unit/test_respond_stage_segmented_reply.py
@@ -2,8 +2,10 @@
 
 With segmented reply enabled, every component used to be sent as its own
 message, so an inline Face in the middle of a sentence split the text into
-several bubbles. The stage now groups consecutive inline components
-(Plain / Face) so a sentence stays in one bubble.
+several bubbles. The stage now attaches a Face to the preceding text bubble
+(or to the following Plain when the chain starts with a Face) while adjacent
+Plain components from the segmentation-words feature (#3959) keep their
+deliberate split.
 """
 
 import math
@@ -14,7 +16,7 @@ import astrbot.core.message.components as Comp
 from astrbot.core.pipeline.respond.stage import RespondStage
 
 
-def test_inline_face_stays_in_the_same_bubble_as_text():
+def test_inline_face_rides_with_the_preceding_text_bubble():
     stage = RespondStage()
     chain = [
         Comp.Plain(text="好的"),
@@ -24,8 +26,27 @@ def test_inline_face_stays_in_the_same_bubble_as_text():
 
     segments = stage._group_segment_chain(chain)
 
-    assert len(segments) == 1
-    assert segments[0] == chain
+    assert segments == [[chain[0], chain[1]], [chain[2]]]
+
+
+def test_leading_face_joins_the_following_text_bubble():
+    stage = RespondStage()
+    face = Comp.Face(id=277)
+    chain = [face, Comp.Plain(text="你好呀")]
+
+    segments = stage._group_segment_chain(chain)
+
+    assert segments == [[face, chain[1]]]
+
+
+def test_adjacent_plain_components_keep_separate_bubbles():
+    stage = RespondStage()
+    first = Comp.Plain(text="第一段。")
+    second = Comp.Plain(text="第二段。")
+
+    segments = stage._group_segment_chain([first, second])
+
+    assert segments == [[first], [second]]
 
 
 def test_components_that_need_separate_sending_stay_alone():
@@ -36,6 +57,17 @@ def test_components_that_need_separate_sending_stay_alone():
     segments = stage._group_segment_chain(chain)
 
     assert segments == [[chain[0]], [record], [chain[2]]]
+
+
+def test_face_after_media_stays_its_own_bubble():
+    stage = RespondStage()
+    record = Comp.Record(file="file:///tmp/a.wav")
+    face = Comp.Face(id=277)
+    chain = [Comp.Plain(text="听"), record, face]
+
+    segments = stage._group_segment_chain(chain)
+
+    assert segments == [[chain[0]], [record], [face]]
 
 
 @pytest.mark.asyncio
--- a/tests/unit/test_respond_stage_segmented_reply.py
+++ b/tests/unit/test_respond_stage_segmented_reply.py
@@ -2,10 +2,10 @@
 
 With segmented reply enabled, every component used to be sent as its own
 message, so an inline Face in the middle of a sentence split the text into
-several bubbles. The stage now attaches a Face to the preceding text bubble
-(or to the following Plain when the chain starts with a Face) while adjacent
-Plain components from the segmentation-words feature (#3959) keep their
-deliberate split.
+several bubbles. The stage now treats an inline Face as glue: it attaches to
+the preceding text bubble and keeps the following Plain in the same bubble,
+while adjacent Plain components from the segmentation-words feature (#3959)
+keep their deliberate split.
 """
 
 import math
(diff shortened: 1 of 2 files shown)


he-yufeng (commented):
The grouping mechanism works for the reported shape, but including `ComponentType.Plain` in `INLINE_SEGMENT_TYPES` over-merges: any run of adjacent text components now collapses into a single bubble, which silently disables segmented reply for plain-text chains.

Concrete counterexample: the segmentation-words feature from #3959 splits LLM output at configured markers into adjacent `Plain` components precisely so segmented reply can deliver them as separate bubbles with intervals. With this change, `["第一段。", "第二段。"]` (two adjacent `Plain` comps) is one `inline_group` and goes out as a single message, so a feature that exists to split text into separate bubbles stops splitting. The same applies to any chain where the model or an upstream stage produced more than one text component without an emoji between them.

A narrower grouping keeps the reported fix without that regression: `Face` attaches to the adjacent bubble (the preceding one, and to the following one only when the chain starts with a `Face`, so a leading emoji never flies solo), while `Plain` always starts a new bubble. That keeps `["好的", Face(277), "，我知道了！"]` in one bubble, leaves `["第一段。", "第二段。"]` as two, and preserves the segmentation-words behavior. A regression test for the consecutive-`Plain` case would pin it either way; right now the tests only cover chains broken up by a `Face` or `Record`, which is why the over-merge passes.

The interval change (word count over the bubble's joined text) looks right, and `Record` can no longer absorb a following `Face` since the group breaks before non-inline comps.

tests/unit/test_respond_stage_segmented_reply.py
```diff
@@ -0,0 +1,56 @@
...
+
+@pytest.mark.asyncio
+async def test_log_interval_for_a_bubble_uses_total_plain_word_count():
+    stage = RespondStage()
+    stage.interval_method = "log"
+    stage.log_base = 10.0
+
+    # "hello" + "world" -> 2 words -> log10(3) ~= 0.477
+    bubble = await stage._calc_comp_interval(
+        [Comp.Plain(text="hello"), Comp.Face(id=277), Comp.Plain(text="world")],
```
EterUltimate, line 91: The grouped text is `"hello" + "world" = "helloworld"`, but `_word_cnt` counts unspaced ASCII text as **1** word (`len(text.split())`), so the interval is drawn from `[log10(2), log10(2)+0.5] = [0.301, 0.801]` while the test asserts `>= log10(3) = 0.477` — ~35% of runs fail (CI evidence: macOS job 103455275079, `assert 0.47712125471966244 <= 0.30788529321914043`; local repro 9/20 on Windows). Adding a trailing space makes it 2 words and the bounds deterministic:

```suggestion
        [Comp.Plain(text="hello "), Comp.Face(id=277), Comp.Plain(text="world")],
```

EterUltimate (commented):
Found the root cause of the macOS `Run pytest suite` failure (job 103455275079): the new test `test_log_interval_for_a_bubble_uses_total_plain_word_count` is statistically flaky, the implementation itself looks correct.

**Root cause:** `RespondStage._word_cnt` (`astrbot/core/pipeline/respond/stage.py` on master) uses the English branch for all-ASCII text: `word_count = len(text.split())`. The grouped bubble text is `"hello" + "world" = "helloworld"`, so `_word_cnt` returns **1**, not the 2 the test assumes. The interval is therefore `random.uniform(log10(2), log10(2) + 0.5) = [0.301, 0.801]`, while the test asserts `>= log10(3) = 0.477` — any sample below 0.477 fails (~35% of runs).

**Evidence:**
- CI: `Run pytest suite (macos-latest)` job 103455275079 failed with `assert 0.47712125471966244 <= 0.30788529321914043` (0.3079 is inside the actual [0.301, 0.801] range).
- Local repro on Windows: 9 failures / 20 runs of this single test.

**Minimal fix:** add a trailing space so the grouped text is `"hello world"` (2 English words); the assertions then hold deterministically — see the suggestion below. Alternatively use CJK text (e.g. "你好"/"世界"), which takes the char-count branch and yields `log10(5)` bounds.

Shxiao101:
Thanks for the catch — agreed that merging Plain runs silently disabled segmentation-words. Reworked in 2da0ca98:

- Only `Face` is inline now: a Face attaches to the preceding bubble when that bubble ends with text (`Plain`/`Face`), and the Plain following a Face is glued back into the same bubble — so `["好的", Face(277), "，我知道了！"]` goes out as the single bubble `好的[狗头]，我知道了！`, and a Face next to media keeps its own bubble.
- Adjacent Plain components without an emoji between them stay separate — `["第一段。", "第二段。"]` remains two bubbles, pinned by the new `test_adjacent_plain_components_keep_separate_bubbles`, together with tests for the leading-Face, Face-after-media, and glue-stops-at-media shapes.
- `_calc_comp_interval` now counts words per Plain component and sums them over the bubble, which also fixes the flaky macOS test (same root cause EterUltimate found below).

One point I'd like to confirm: the comment mentions both "`Plain` always starts a new bubble" and "keeps `[\"好的\", Face(277), \"，我知道了！\"]` in one bubble", which conflict for a mid-sentence Face — the glue reading avoids a bubble starting with "，". I went with the glue semantics; happy to switch to the strict reading if you prefer.

tests/unit/test_respond_stage_segmented_reply.py
Shxiao101: Good diagnosis, matches what we found. Rather than adjusting the test, the implementation now counts words per Plain component and sums them over the bubble (2da0ca98), so `hello` + `world` = 2 words and the asserted bounds are exact — no trailing-space workaround needed.

kilisamemarisaaa (commented):
Reviewed commit c8fa9633f10306d0f87be518b8095b3a9ec04c39. The revised grouping preserves adjacent Plain components as separate bubbles, joins text across an inline Face, and keeps Record components separate. In the send path, Reply/At headers remain attached to the first eligible segment. Counting words per Plain component also addresses the earlier hello/world interval-test mismatch.

Local Windows/Python 3.12 validation: tests/unit/test_respond_stage_segmented_reply.py passed all 7 cases (one audioop deprecation warning); Ruff check passed for both changed files. This was local unit validation, without live QQ delivery.

The current Windows CI job is still red: job 103526635181 reports 2697 passed and one failure in tests/unit/test_event_loop_diagnostics.py::test_event_loop_watchdog_writes_rotating_log, specifically the captured-stack filename assertion. The segmented-reply tests are not the reported failure. That CI failure still needs to be resolved before treating the whole PR as green.

PR #10055, opened 2026-09-12
fix: avoid blocking audio format probes

zzz27578:
## Summary

- run the synchronous audio magic-byte probe in `asyncio.to_thread` so audio conversion does not block the event loop
- preserve the no-op fast path for Ogg Opus input requested as Ogg
- add regression coverage for the asynchronous probe and Ogg/Opus fast path

## Context

Follow-up to #9329. The current `master` includes the file-content check from #9612, but the probe still performs synchronous file I/O inside `convert_audio_format`, and Ogg Opus files are needlessly re-encoded when Ogg output is requested. This PR keeps the correctness fix on the latest `master` while addressing those remaining cases.

## Validation

Per request, local tests were not run.

## Summary by Sourcery

Keep audio conversion responsive by offloading format probes and preserving compatible Ogg Opus inputs.

Bug Fixes:
- Prevent synchronous audio format detection from blocking the asyncio event loop.
- Avoid unnecessary re-encoding when Ogg Opus input is requested as Ogg.

Tests:
- Add regression coverage for asynchronous audio probing and the Ogg Opus no-op path.

### DIFF
--- a/astrbot/core/utils/media_utils.py
+++ b/astrbot/core/utils/media_utils.py
@@ -1135,10 +1135,17 @@ async def convert_audio_format(
         Exception: Raised when ffmpeg is unavailable or conversion fails.
     """
     source_path = Path(audio_path)
-    if source_path.suffix.lower() == f".{output_format}" and (
-        not source_path.exists() or _get_audio_magic_type(audio_path) == output_format
-    ):
-        return audio_path
+    if source_path.suffix.lower() == f".{output_format}":
+        # The magic-byte probe performs synchronous file I/O. Keep it off the
+        # event loop because this helper is called while processing messages.
+        if not source_path.exists():
+            return audio_path
+
+        detected_format = await asyncio.to_thread(_get_audio_magic_type, audio_path)
+        if detected_format == output_format or (
+            output_format == "ogg" and detected_format == "opus"
+        ):
+            return audio_path
 
     if output_path is None:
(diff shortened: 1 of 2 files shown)


kilisamemarisaaa (commented):
Independent validation at e87159966c796176961c22388e8221dcfa03d17e on Windows:

- `pytest tests/test_media_utils.py -q`: 55 passed (one audioop deprecation warning).
- Ruff check and format --check on both changed files: passed.
- Added a local diagnostic wrapper around the real `_get_audio_magic_type` recording `threading.get_ident()`, without replacing `asyncio.to_thread`. The off-loop-thread assertion fails on base 7ec39bdef02a55964adf8f2b3fb4fa2b53c8bab6 and the off-loop-thread assertion passes on this head (1 test). The wrapper still invokes the original detector on a WAV fixture.

This confirms actual thread offloading in addition to the new mocked-call unit test. The existing Ogg/Opus no-subprocess regression also passes as part of the 55 tests. Scope: media utility tests and the probe thread boundary, not a full-suite/real-codec integration run or a claim that every filesystem operation in the helper is asynchronous.

zzz27578:
@kilisamemarisaaa Thanks for the independent validation. The targeted media tests and Ruff checks passing, together with the real thread-ID check, confirm that the magic-byte probe is actually offloaded from the event-loop thread and that the Ogg/Opus fast path remains intact. No code changes are needed from this review.

PR #10067, opened 2026-09-13
fix(qqofficial): finalize C2C stream to prevent replies reverting to the first fragment

Linyesantan:
## 问题

QQ 官方机器人（WebSocket，C2C 私聊）开启流式输出后，回复会先正常流式增长为完整内容，随后**回滚成开头的几个字**；流式消息有时还会长期停在“生成中…”。

真实案例：会话数据库 `data_v4.db` 里保存的是完整 **850 字** 回复，但 QQ 客户端里那条流式消息最终只剩首包 **5 个字**，且全程没有任何报错日志。

关联 issue：#10066

## 根因

`send_streaming` 每次节流发送后都会清空 `self.send_buffer`。正常结束时执行的 `state=10` 收尾分支，如果 buffer 已被最后一次节流发送清空，`_post_send` 会因为 `if not self.send_buffer: return None` 直接返回——**结束帧从未发出**，QQ 侧流式消息始终没有 finalize，超时后回滚 / 停留在未完成状态。异常、断流路径同样会丢弃尚未发送的尾段。

## 改动

- 追踪 `full_text` / `sent_len`：正常结束时把未下发的尾段以 `state=10` 补齐；若尾段为空则发送 `"\n"` 收尾帧（满足 QQ API 结尾换行要求，且不会被 `_post_send` 的“空 buffer 早退”丢弃）。
- 断流 / 异常路径同样把未下发尾段以 `state=10` 补发，避免消息冻结在最后一个分片。
- 若中途分片从未拿到 stream id，则降级为普通消息补发全文。
- 新增回归测试 `tests/test_qqofficial_stream_finalize.py`。

## 与 #9875 的关系

本 PR 包含 #9875 的修复逻辑（作者 @VZService-AI，commit `bb19453`），并按 #5264 的要求补上了回归测试。若 #9875 先行补测试并合并，本 PR 可关闭。

## Test plan

- [x] 新增回归测试：`pytest tests/test_qqofficial_stream_finalize.py` → 3 passed
  - 节流发送已吃掉全部内容时，仍必须发出 `state=10` 结束帧；
  - 存在未下发尾段时，最终帧只携带该尾段（不重复）；
  - 上游中途断流时，未下发尾段仍以 `state=10` 补发。
- [ ] 真实 QQ 官方 C2C 环境验证长回复不再回滚。

## Summary by Sourcery

Ensure QQ Official C2C streaming replies are finalized reliably and preserve complete generated content across normal and interrupted delivery paths.

Bug Fixes:
- Finalize QQ Official C2C streaming replies so completed messages no longer revert to an initial fragment or remain stuck in a generating state.
- Flush unsent content when streaming ends unexpectedly, and fall back to sending the full response as a regular message when no stream ID is available.

Tests:
- Add regression coverage for finalizing fully flushed streams, sending only the unsent tail, and recovering remaining content after upstream interruption.


kilisamemarisaaa (approved):
Reviewed at head 7c7a0e2447821ec72c4ab3b73b285d65944fdcdb. The fix preserves the generated Plain text and sent length across throttled C2C sends, emits a non-empty state=10 frame even when the buffer was fully flushed, flushes only the unsent tail on normal completion or upstream failure, resets tracking at tool breaks, and falls back to a regular full-text message when no stream id is available. I also checked the existing stream-buffer-copy coverage. Windows/Python 3.12 focused tests: tests/test_qqofficial_stream_finalize.py and tests/test_qqofficial_stream_buffer_copy.py passed 11 tests; Ruff and git diff --check pass. No correctness blocker found in the non-live-QQ paths.

kilisamemarisaaa (changes requested):
Follow-up after reviewing the full current review set: I am changing my review to CHANGES_REQUESTED because two correctness gaps remain at head 7c7a0e2447821ec72c4ab3b73b285d65944fdcdb. (1) In the tool-break branch, if self.send_buffer skips finalization after a throttled state=1 send has already cleared the buffer; when stream_payload.id exists, the active QQ stream still needs a state=10 tail/newline frame before resetting the payload. Please add a regression test for a break immediately after a flushed fragment. (2) If a state=1 request has been accepted but returns no extractable id, the final branch sends ull_text as an ordinary message. That can duplicate the prefix already delivered by the stream. Please define a non-duplicating fallback (or only fall back when no partial stream was delivered) and test the no-id-after-partial-send path. The existing 11 focused tests pass, but they do not cover these two branches.

PR #9303, opened 2026-07-17
fix(kb): clean up documents and media when deleting a knowledge base

he-yufeng:
`KnowledgeBaseManager.delete_kb()` deletes the `KnowledgeBase` row and the vector DB, but the KB's `KBDocument` and `KBMedia` rows carry `kb_id` with no ORM relationship or foreign-key cascade. Deleting a knowledge base therefore leaves **all** of its documents and media orphaned in the SQLite database.

This is the same cascade gap #9120 fixed for single-document deletion (`delete_document_by_id` deletes `KBMedia` explicitly because a bulk `DELETE` does not trigger ORM cascade) — the whole-KB delete path was the missed sibling.

删除知识库时只删了 `KnowledgeBase` 行和向量库，但该库的 `KBDocument` / `KBMedia` 通过 `kb_id` 关联、没有 ORM/外键级联，会把全部文档和多媒体记录留成孤儿（与 #9120 修的单文档删除是同一类问题）。

### Modifications / 改动点

- Add `KBSQLiteDatabase.delete_kb_by_id(kb_id)` that bulk-deletes `KBMedia`, `KBDocument`, and the `KnowledgeBase` row for the KB, mirroring `delete_document_by_id`.
- `delete_kb()` calls it (after `delete_vec_db()`) instead of only `session.delete(kb_helper.kb)`.
- Unit tests in `tests/unit/test_kb_document_cleanup.py`: whole-KB deletion leaves no orphaned document/media, and deleting one KB does not touch another KB's rows.

- [x] This is NOT a breaking change. / 这不是一个破坏性变更。

### Screenshots or Test Results / 运行截图或测试结果

```
$ .venv/bin/python -m pytest tests/unit/test_kb_document_cleanup.py -q
6 passed
```

Red-before-green check: with the two child-table deletes removed from `delete_kb_by_id`, `test_delete_kb_cleans_documents_and_media` fails because the `KBDocument`/`KBMedia` rows survive the KB deletion; with the fix in place they are gone.

### Checklist / 检查清单

- [x] 👀 Well-tested (unit tests + red-before-green verification).
- [x] 🤓 No new dependencies.
- [x] 😮 No malicious code.

## Summary by Sourcery

Ensure deleting a knowledge base also removes its associated documents and media, avoiding orphaned rows in SQLite.

Bug Fixes:
- Fix knowledge base deletion leaving orphaned KBDocument and KBMedia records in the SQLite database.
[...]

PR #10072, opened 2026-09-14
fix: keep failed knowledge-base document deletions retryable

beemines:
Deleting a knowledge-base document currently commits removal of its `KBDocument` and `KBMedia` rows before deleting its chunks and FAISS vectors. If index persistence or chunk deletion then fails, the request raises an error but the document has already disappeared from the document list. The remaining stored chunks have lost their metadata and the user cannot select the document in the dashboard to retry deletion.

### Modifications

Move `vec_db.delete_documents()` before the metadata transaction. A storage failure keeps the document and media records discoverable; retrying completes the existing idempotent deletion. A later metadata-transaction failure also retains these records for retry, even when the chunks have already been removed.

This is an ordering fix in the knowledge-base application workflow. It does not make the separate stores atomic, restore vectors after a partial failure, add automatic retries, or change FAISS internals. API schemas and normal successful deletion are unchanged.

Add parameterized tests using real temporary SQLite metadata/chunk databases and a real FAISS index. The embedding provider returns a fixed vector; faults are injected at index save, chunk deletion, and the document-row deletion inside the metadata transaction. The tests reopen the metadata database, retry deletion, check the other document's data remains, and reopen the vector store to verify persistence.

This was found by inspecting the current deletion path. Related #9120 handles media cleanup on successful single-document deletion; #9303 handles whole-knowledge-base cascades. Neither changes this single-document failure ordering.

### Verification steps and test results

Windows, Python 3.12; base `bd046ed29914ee559e9bf47676ccb71a84f747ba`.

Before the production change, the index-save and chunk-delete regression cases both failed because the target document was absent after reopening the metadata database:

```text
2 failed, 1 warning
```

After the fix, including the metadata rollback/retry case:

```text
python -m pytest tests/unit/test_kb_delete_retry.py tests/unit/test_kb_document_cleanup.py tests/unit/test_kb_upload_atomicity.py tests/unit/test_kb_manager_resilience.py tests/unit/test_knowledge_base_service_contract.py tests/unit/test_faiss_vec_db.py tests/unit/test_document_storage_fts.py -q
56 passed, 3 warnings
```

The warnings include the existing `audioop` deprecation and aiosqlite worker-thread `Event loop is closed` warnings. They are not failed assertions. The full repository suite, Linux/macOS runs, live model calls, and browser-level interactions were not run locally.

`ruff format --check .`, `ruff check .`, and `git diff --cached --check` pass. The new test file was also explicitly formatted and linted.

### Checklist

- New-feature discussion: not applicable; this is a bug fix.
- [x] Verification steps and actual test output are provided above.
- [x] No new dependencies.
- [x] No malicious code.

### AI assistance

Codex assisted with code inspection, reproduction, implementation, validation, and this PR description. The tests and results above were executed locally; the change does not claim end-to-end atomicity across independent stores.

### CI follow-up: synchronize watchdog tests

The first remote run passed on Ubuntu and Windows. On macOS, all three new knowledge-base tests passed, but `test_event_loop_watchdog_writes_rotating_log` failed because the sampled stack did not include the test coroutine (2886 passed, 1 failed). Its fixed 50ms blocking window can end before log opening and stack capture complete.

A separate test-only commit replaces the fixed sleeps in the two watchdog dump tests with bounded `threading.Event` waits and `finally` cleanup. The rotation test still writes real stack traces and checks the original rotated file contents; it now specifically checks the event-loop thread's stack contains the test coroutine. Production watchdog behavior is unchanged.

Local fault injection adding 80ms to the real log-open operation reproduces the original missing-test-frame assertion; the same probe passes with synchronization. The watchdog module plus the new knowledge-base tests pass together (`8 passed`), and full Ruff formatting/lint checks pass. Remote CI on commit `c3102950ffd4b919bf1a206e522a5dfbcfdb6dcc` is now complete: all 23 checks passed. The actual test steps on Ubuntu, Windows, and macOS all succeeded. Formatting, CodeQL, dashboard CI, and smoke tests also passed. [Unit-test run](https://github.com/AstrBotDevs/AstrBot/actions/runs/34795561009).

## Summary by Sourcery

Make single-document knowledge-base deletion retryable by removing stored chunks before committing document metadata removal.

Bug Fixes:
- Keep knowledge-base documents and their media metadata discoverable when vector, chunk, or metadata deletion fails, allowing safe retry of the deletion.

CI:
- Synchronize watchdog tests with event-based waits to eliminate platform-dependent timing failures; all CI checks now pass.

Tests:
- Add parameterized integration-style tests covering failures at index persistence, chunk deletion, and metadata deletion, including retry behavior and persistence across database restarts.


kilisamemarisaaa (approved):
Reviewed at head c3102950ffd4b919bf1a206e522a5dfbcfdb6dcc. The deletion order now removes vector/chunk storage before the SQLite transaction, so failures at index save, chunk deletion, or metadata deletion leave the document discoverable for an idempotent retry. The new regression suite exercises all three failure boundaries and restart persistence; local Windows/Python 3.12 run: 56 passed (7 pre-existing warnings). All repository CI checks are green, including Windows/macOS/Linux pytest, Python 3.10-3.14 smoke tests, CodeQL, format-check, and build.

PR #10087, opened 2026-09-14
feat: support explicit reply IDs in message sending tool

PeiPei233:
The agent can send messages but cannot explicitly quote an earlier message, even though message chains already support `Reply`. Add an optional `reply_to_message_id` argument to `send_message_to_user` so callers can quote a known platform message ID in the target session.

### Modifications

- Expose the optional string argument and document when to use the tool for an explicit reply. The description requires a real target-session platform message ID, not a history database row ID.
- Validate non-empty string IDs and prepend a `Reply` component to the outgoing chain. Omitted arguments retain existing behavior and cross-session permission checks remain in place.
- Reuse platform adapters' existing reply support; this does not add support to adapters that do not handle replies, or verify that a message exists on the platform.

Example tool arguments:
```json
{"messages":[{"type":"plain","text":"About your earlier message..."}],"reply_to_message_id":"123456789"}
```

### Test Results

- `python -m pytest tests/unit/test_message_tools.py tests/unit/test_aiocqhttp_reply.py -q`: 32 passed.
- Covers explicit reply ordering and OneBot serialization, current and explicit target sessions, invalid IDs, omitted IDs, and cross-session authorization.
- `ruff format .` and `ruff check .`: passed.
- `git diff --check`: passed.

### Checklist

- [ ] New feature discussed with maintainers in advance.
- [x] Changes tested with verification steps and results above.
- [x] No new dependencies introduced.
- [x] No malicious code introduced.

## Summary by Sourcery

Enable callers to quote a known earlier platform message when sending messages without changing existing sending behavior.

New Features:
- Add optional explicit reply targeting to the message-sending tool using platform message IDs.

Enhancements:
- Preserve existing message behavior and cross-session authorization while supporting reply component ordering and platform serialization.

Documentation:
- Document when to use explicit reply targeting and distinguish platform message IDs from history database row IDs.

Tests:
- Add coverage for explicit replies, target sessions, invalid and omitted IDs, serialization, and permission enforcement.


kilisamemarisaaa (approved):
Reviewed head 37c97968af95a113ff2cc17f3cfa497d0008ef5a. The optional reply_to_message_id is validated as a non-empty string before constructing the outgoing chain, the Reply component is placed before user content, and the existing target-session authorization path remains unchanged. The tests cover current and explicit sessions, OneBot serialization, invalid values, omitted values, and member cross-session denial. All required Linux/Windows/macOS pytest lanes, CodeQL, Ruff/format, build, and smoke checks are green. I found no correctness blocker.

PR #10098, opened 2026-09-15
fix: resolve each OneBot file segment by its own ID

xiongyuyang:
### Modifications

The OneBot message converter reads the first message segment's file ID when resolving every NapCat attachment. A leading text segment therefore causes attachments to be skipped with KeyError; multiple file segments repeatedly resolve the first file.

Read the current file segment's ID for both group and private messages. Add regression coverage for two attachments with and without leading text in each message type.

Verification: `python -m pytest tests/unit/test_aiocqhttp_files.py tests/unit/test_aiocqhttp_reply.py tests/unit/test_aiocqhttp_poke.py tests/unit/test_aiocqhttp_group_info.py -q -p no:cacheprovider`

Before the fix: all four new regression cases failed (repeated first ID or no URL resolution).
After the fix: 19 passed, 1 existing audioop deprecation warning on Windows/Python 3.12.
Ruff check and commit hooks passed. Tests exercise the real message converter with mocked OneBot API responses; no live QQ integration was run.

### Checklist

- [x] Changes are tested and verification steps/results are provided above.
- [x] No new dependencies.
- [x] No malicious code.

## Summary by Sourcery

Fix OneBot attachment URL resolution so every file segment is converted independently.

Bug Fixes:
- Resolve each OneBot file segment using its own file ID for group and private message attachments, including messages with leading text.

Tests:
- Add regression coverage for multiple file attachments in group and private messages with and without leading text.


kilisamemarisaaa (approved):
Reviewed at head 42b0afa5e24d1afd48cf5dabe8593088a307b63a. The converter now uses each current file segment's file_id for both group and private OneBot actions, so leading text no longer shifts lookup and multiple attachments no longer reuse the first ID. In a Windows/Python 3.12 tarball checkout of this exact head, the author's focused suite passed 19 tests (4 pre-existing warnings); Ruff passed. The regression matrix covers group/private messages, leading text, and two distinct file attachments with mocked API responses. No correctness blocker found.

Soulter (approved):

PR #10099, opened 2026-09-15
fix: preserve case in GPT-SoVITS synthesis parameters

xiongyuyang:
### Modifications

GPT-SoVITS default parameters currently lowercase every value. For example, `/Voices/Alice.WAV` becomes `/voices/alice.wav`, which can break reference audio lookup on case-sensitive filesystems. Prompt text such as `Hello NASA` is also modified before sending it to the server.

Preserve string casing while retaining string serialization for numeric values and lowercase `true`/`false` for booleans. Add parameterized regression coverage for reference audio paths, prompt text, both boolean values, integers and floats.

Verification: `python -m pytest tests/unit/test_gsv_selfhosted_source.py -q -p no:cacheprovider`

Before the fix: 2 failed (path and prompt casing), 4 passed.
After the fix: 6 passed, 1 existing audioop deprecation warning on Windows/Python 3.12.
Ruff check and commit hooks passed. These tests exercise request parameter construction; a running GPT-SoVITS server/model was not used.

### Checklist

- [x] Changes are tested and verification steps/results are provided above.
- [x] No new dependencies.
- [x] No malicious code.

## Summary by Sourcery

Preserve GPT-SoVITS synthesis parameter values without altering user-provided string casing.

Bug Fixes:
- Preserve the casing of GPT-SoVITS string synthesis parameters while continuing to serialize booleans as lowercase and numeric values as strings.

Tests:
- Add parameterized regression coverage for reference audio paths, prompt text, booleans, integers, and floats.


kilisamemarisaaa (approved):
Verified commit e102780a40f638879deb5a9c077809e3bdc9914a locally from the PR tarball on Windows/Python 3.12: python -m pytest tests/unit/test_gsv_selfhosted_source.py -q passed (6 passed; only the existing audioop deprecation warning). Ruff check passed for both changed files. The implementation preserves string casing, serializes numeric values as strings, and keeps booleans lowercase for the GPT-SoVITS API. No additional issues found.

PR #10100, opened 2026-09-15
fix: bind CosyVoice request authentication to its provider

xiongyuyang:
### Modifications

Creating a second DashScope TTS provider overwrites `dashscope.api_key`. A later CosyVoice request from the first provider then sends the second provider's bearer token, potentially causing authentication failures or charging the wrong account.

Set an explicit per-request Authorization header from the selected provider's key. Remove case-insensitive custom Authorization entries before adding it, avoiding duplicate authentication headers. Retain SDK initialization compatibility: the installed SpeechSynthesizer still requires the global key during construction, but its public headers argument overrides the generated bearer header for the outgoing request. Qwen already receives its key explicitly.

`python -m pytest tests/unit/test_dashscope_tts_auth.py tests/test_dashscope_embedding_source.py -q -p no:cacheprovider`

22 passed, 2 dependency deprecation warnings on Windows/Python 3.12. The original two-provider regression failed before the fix. Tests inspect headers generated by the real SDK, mock only synthesis execution, exercise both providers concurrently and cover custom header casing plus Qwen's existing key isolation. No paid API was called.

Ruff check and commit hooks passed.

### Checklist

- [x] Tested, with verification steps and results above.
- [x] No new dependencies.
- [x] No malicious code.

## Summary by Sourcery

Ensure DashScope TTS requests consistently authenticate with the API key of the provider that issued them.

Bug Fixes:
- Bind CosyVoice authentication to the selected provider’s API key for every request, preventing cross-provider credential leakage and incorrect account usage.
- Remove existing Authorization headers before applying the provider-specific bearer token to ensure a single effective authentication header.

Tests:
- Add coverage for concurrent CosyVoice providers, custom Authorization header casing, and Qwen credential isolation.


kilisamemarisaaa (approved):
Reviewed at head d32ff9a1998991a8520bf9018a91160cb2a3d9fb. The CosyVoice path removes all case variants of a caller-supplied Authorization header before adding the selected provider's bearer token, while preserving other custom headers and the SDK's construction-time global-key compatibility. The exact-head Windows/Python 3.12 focused suite passed 22 tests (5 dependency warnings); Ruff passed. The tests inspect headers produced by the real SDK and cover concurrent providers, stale custom header casing, and Qwen's existing per-provider key path. No credential-isolation blocker found.

PR #10182, opened 2026-09-21
fix: coalesce queued private chat replies

94yi:
## Summary by Sourcery

Coalesce eligible queued private messages into one timely response while preventing stale or terminated requests from reaching the LLM.

New Features:
- Coalesce eligible queued private text messages from the same conversation into a single bounded reply prompt.

Bug Fixes:
- Prevent stale queued messages, superseded requests, and session termination from triggering redundant or delayed LLM replies.

Enhancements:
- Preserve separate handling for commands, non-text messages, and messages from different private conversations while retaining receive order in merged prompts.

Tests:
- Add regression coverage for same-session coalescing, session isolation, stale backlog dropping, command exclusion, and termination before LLM requests.


DFGHJ43:
你好！你做出的修改已经属于新功能而不是修复的范畴了，添加新功能之前最好在issue和其他开发者讨论一下再添加，而且其实已经有社区插件实现了合并临近消息的功能了，你可以看一看。

